PR #4, opened 2017-02-17
Tapeout

shunshou:
lmk if you want anything changed or feel free to just push to this yourself.


shunshou:
Ok so I should have src/main/scala/barstools/tapeout and src/test/scala/barstools/tapeout? And are you ok w/ package name or would you prefer other things? I'll change and then merge.

colinschmidt:
Yeah thats the correct directory structure.

Another package name I had thought of was barstools.firrtl but I think it depends on how we imagine the content of this repo changing. If we think there will be non-tapeout stuff we would want to have something like barstools.util as well as barstools.tapeout.

I think overall I would change it to barstools.tapeout.firrtl so we don't ever have to move the tapeout specific firrtl passes. And we may or may not end up fleshing out the remaining packages but I think thats ok.

shunshou:
I guess the wonky thing is that all of these passes are half Firrtl and half Chisel (since the ResetInverter trait is more of a Chisel thing). barstools.tapeout.transforms maybe?

colinschmidt:
Oh yeah I forgot about that. I think the transforms package

shunshou:
Hmm if you're ok w/ these naming conventions, care to merge? ;)

colinschmidt:
@jackkoenig Has something to say about the build.sbt. I'll give him a sec to comment.

build.sbt
```diff
@@ -1,16 +1,31 @@
...
 
-lazy val executionoptions = (project in file("executionoptions"))
-  .settings(commonSettings)
-  .settings(
-    libraryDependencies ++= executionoptionsDependencies
-  )
+libraryDependencies ++= Seq(
+  "org.scalatest" %% "scalatest" % "2.2.5",
+  "org.scalacheck" %% "scalacheck" % "1.12.4"
+)
```
jackkoenig: While this is probably okay right now, we're going to need to make this a multi-project build in the near future. We would like to factor out some common utils from Chisel and Firrtl here so they can both depend on it (which means there needs to be a subproject for things that depend on chisel3 and a different subproject for things that don't). As long as we don't publish this globally it's fine, but basically we probably need a more specific project name for these things that depend on chisel3.

shunshou: I assume that's kind of independent to the specific libraryDependencies code you're referring to? Is it better to do something like in chisel3 aka:

```
lazy val coreMacros = (project in file("coreMacros")).
  settings(commonSettings: _*).
  settings(publishArtifact := false)
```

? And if so, should the directory structure not be src/main/scala/barstools/blah but instead tapeout/src/main/scala/xyz?

shunshou:
@jackkoenig @colinschmidt how's that?

jackkoenig:
Looks good to me!

PR #5, opened 2017-02-18
Move passes from pfpmp to barstools.

colinschmidt:
Simply include the pfpmp passes for now. Integrated with hurricane-chip and tested there.

More work to be done but some needs extra features and the rest can be done incrementally as tapeout schedule permits.

### DIFF
--- a/tapeout/src/main/scala/transforms/ConvertToExtModPass.scala
+++ b/tapeout/src/main/scala/transforms/ConvertToExtModPass.scala
@@ -1,0 +1,36 @@
+// See LICENSE for license details.
+
+package barstools.tapeout.transforms
+
+import firrtl._
+import firrtl.ir._
+import firrtl.passes.Pass
+
+// Converts some modules to external modules, based on a given function.  If
+// that function returns "true" then the module is converted into an ExtModule,
+// otherwise it's left alone.
+class ConvertToExtModPass(classify: (Module) => Boolean) extends Pass {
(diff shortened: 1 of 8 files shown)


shunshou:
@colinschmidt any idea if there's some better way to generalize reading Yaml files? Seems like between Annotations, MemLib stuff, your stuff, and the pad pass I'm working on, there's just lots of different Yaml readers... that are 90% the same, save for implicits. :\

colinschmidt:
I don't think I've added YA-YAML reader, I was trying to just reuse the one that chisel/firrtl use for annotations. I also haven't worked with YAML much at all, so I think we should ask @chick @azidar if they have.
But I do agree it would be nice to somehow standardize on this. I actually am just vaguely remembering a conversation with @jackkoenig on Friday about datatypes in annotations which may or may not be related.

shunshou:
Btw - is the plan to have multiple branches? B/c if not, you're free to merge (unless we're hoping to get some more legit code review from people... but for the sake of tapeout time constraints, I'm kind of meh about that >_>...).

colinschmidt:
I think its better to have one branch. Will merge later today.

shunshou:
Okiedokes. I have my own branch going, but I don't expect any conflict. Still experimenting with templating, etc.

colinschmidt:
As to YAML parsing see ucb-bar/firrtl#452

PR #7, opened 2017-02-19
Add Pads + other utilities

shunshou:


shunshou:
@colinschmidt when I finish cleaning this up a little more, can you take a look at my methodology and see if this makes sense for future tapeouts + maybe want more features? (prob too late to use for Hurricane, esp. since I assume you guys have your own create_pads.py thing going).

colinschmidt:
Yes, of course.

Yeah we are still using python :(

shunshou:
@colinschmidt lmk what you think of this? Relevant examples for pad transform users:
https://github.com/ucb-bar/barstools/blob/20726bcd51faf3bd0e7b4f978d42ee3f467863d5/tapeout/src/main/resources/IOPads.yaml
https://github.com/ucb-bar/barstools/blob/20726bcd51faf3bd0e7b4f978d42ee3f467863d5/tapeout/src/test/scala/transforms/pads/AddIOPadsSpec.scala

You specify a wrapper for the pad with the Yaml config file. I'm hoping this is general enough to support most foundry pads. And a use case for how you add pads (by specifying that a module is a "TopModule" + annotatePad) is in the spec. Specifying a pad name is optional--if it's not found, it'll match type i.e. GroundType -> Digital in/out AnalogType -> Analog. By specifying a pad name, you can have multiple kinds of (custom) digital, analog pads. You can also say that a port doesn't have a pad associated with it (in case you want to directly connect up some blackbox to a bump). You need to specify a pad side; otherwise a default side will be assigned. 

Todo: Figure out how to guarantee order of passes (black box helper needs to be done AFTER pad pass -- it seems to do that in my test case, but I'm not sure how generalized this is, and I'd like not to have to specify annotations directly in the module itself...). I also want to be able to add supply pads, as in create_pads.py and specify some kind of pad config file for floor planning. 

I'm using Handlebars to feed in Verilog,Tcl, etc. templates via a Yaml file that then get populated. 

@azidar if you have any comments on the pass itself, and things aren't too painful to rewrite... I'd appreciate that. Maybe it'd give you more insight on how outsiders can write Firrtl passes. 

General principle:
Make pad annotation in Chisel
Handle all annotations
For PadTemplate annotation, collect usable pads from the Yaml file
For the Top = Pad module, collect ports that need pads
Create parameterized blackboxes (that also do bit extraction/cat with a wrapper on top of the Yaml Verilog) for the pads. 
Hook up the pads: NewTop (= new circuit.main, named the same as old Top) has 2 submodules = Old Top and PadFrame.

shunshou:
Also, @stevobailey @grebe if you guy don't already have a create_pads.py going, might be worth giving this a shot and seeing if it meets any of your needs.

shunshou:
[NOTE: Major refactor. Don't bother code reviewing yet.]

colinschmidt:
Is the major refactor done now? i.e. Worth checking out add pads?

shunshou:
Ah yessir. I'm also adding some clk stuff in, but pads stuff preliminarily is done.

colinschmidt:
Related to #10 @shunshou how would you feel about merging https://github.com/ucb-bar/barstools/pull/7/commits/654dac0b796b04e843caa2997a9d664982380c3f into master? I think that puts everything but the clock stuff on master?

shunshou:
I changed some stuff about add_pads in recent commits. I would just merge
the latest add_pads in. I'm prob not gonna finish clk stuff for the next
several weeks (need to actually write hardware now), but it doesn't break
any existing functionality.

colinschmidt:
Alright go for it

PR #10, opened 2017-03-06
Merge master into barstools

stevobailey:
I'd like to use the new passes in master for CRAFT, but they aren't merged into add_pads yet. Is there anything stopping this from being okay?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,5 @@
+# barstools
+Useful utilities for BAR projects
+
+Passes/Transforms that could be useful if added here:
+* Check that a module was de-duplicated. Useful for MIM CAD flows and currently done in python.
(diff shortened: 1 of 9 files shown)


colinschmidt:
I think we had wanted it to go the other way. We should merge the add pads stuff into here, we can do that now if @shunshou is ok with it. I'd rather merge #7 rather than this.

stevobailey:
Right, but we're on the add_pads branch right now, though an older commit. I think commit 		da59a73 is ready for merge, but @shunshou has added other stuff since then. I'd be fine if we merge #7, just so long as it builds.

shunshou:
I think my stuff should build (also made some changes that would make some
of brians add_pads work non-compliant, but would be a quick fix). I'll
merge add_pads with master at hopefully around 9 if that's ok? I'm out
until then (and also not back in Berkeley until 9pm tmrw).

stevobailey:
I'll just do it now and bug you if I run into issues. Thanks!

shunshou:
If brians add pads stuff is in your version of craft chip, I think the
compile time warnings should be enough to show you what's going on. Made
some things less boiler plate-y mostly.

PR #11, opened 2017-03-06
[stevo]: combine generates, make it a trait

stevobailey:
As the title says. Should be backwards compatible with any existing code. It compiles, but haven't tested it. Perhaps some hurricane folks could?

Also, I've made this a trait so I can extend it in my own Generate (and so can you!).


stevobailey:
Had to make it lazy for things to work, but now I've tested it.

shunshou:
Definitely ask @colinschmidt to review this. I think separating generates helps with clarity. Not sure if it makes sense to have "App" as a trait....

stevobailey:
I was following the rocket-chip style of Apps as traits. The separate generates just had a lot of code redundancy.

colinschmidt:
What version of firrtl and chisel3 is this supposed to work with?

stevobailey:
It's working with our versions:

Chisel3:
commit 3e6ef13ff5cda2e65efbbf5d306cc101582ad0e5
Author: Siddhanathan Shanmugam <siddhanathan+social@gmail.com>
Date:   Tue Feb 28 17:58:42 2017 -0800

    Fix minor typo in readme (#537)

Firrtl:
commit e8a8fd794cbddaaaa1dbdb1fdb10b8bb46c06ffe
Author: Colin Schmidt <colins@berkeley.edu>
Date:   Tue Feb 28 11:44:11 2017 -0800

    Fix validation print for log-level (#394)

edwardcwang:
@stevobailey Does it this work with chisel3 and firrtl master?

stevobailey:
Probably, though I haven't explicitly tested it against anything beyond the versions I mentioned above.

colinschmidt:
It was difficult to review this without making lots of changes, so I just made them myself. @stevobailey what do you think of the changes.

colinschmidt:
Worth noting that despite deleting files this PR before actually increased the lines of code (it did add some safety) but with these cleanups we have more functionality and fewer lines of code :)

stevobailey:
This builds against our code, so I'm okay with this. The big idea was to reduce the file count, generalize the code, and make it a trait so other Generators could extend it. These requirements are met. I kept the object names the same so other projects (e.g. hurricane) don't need to change Makefile code when this repo is updated.

colinschmidt:
I agree there was a lot of duplication before, and although some people argue that more small files are better it would be difficult to do in this context.
I also think keeping the object names the same makes sense.
No code changes in hurricane were required for this.
I'm going to squash and merge, after class.

PR #12, opened 2017-03-12
Add firrtl override option and quick start example

edwardcwang:


chick:
@edwardcwang any thoughts on what should be done with this?

edwardcwang:
We can fix it if you think it's useful; otherwise just close it.

PR #13, opened 2017-03-13
Fix pad/clkgen transforms for firrtl/#459 issue.

azidar:
Should all work now.


shunshou:
Just to double check, but, if the end of my pads transform does:

```
        val cs = CircuitState(
          runPasses(circuitWithBBs, passSeq), 
          LowForm, 
          Some(AnnotationMap(prevAnnos ++ bbAnnotations))
        )
        // TODO: *.f file is overwritten on subsequent executions, but it doesn't seem to be used anywhere?
        (new firrtl.transforms.BlackBoxSourceHelper).execute(cs)
```

B/c I propagate old + new annotations to the BlackBoxSourceHelper transform, things should be ok?

PR #16, opened 2017-03-15
Update README.md with example invocation

azidar:


colinschmidt:
I think this is fine for now but we should perhaps include a rocket-chip example as a lot of people will end up interacting with barstools that way.

Here is my current call to sbt
```
$(SBT) "project barstools" "run-main barstools.tapeout.transforms.GenerateTopAndHarness -i $(abspath $(filter %.fir,$^)) --top-o $(generated_dir)/$(long_name).top.v --harness-o $(generated_dir)/$(long_name).harness.v --anno-file $(abspath $(generated_dir)/$(long_name).anno) --syn-top $(VLSITOP) --harness-top $(MODEL) --seq-mem-flags \"-i:$(vlsi_dir)/src/sram_pins.$(VLSITOP).yaml:-o:$(generated_dir)/$(long_name).conf\" --list-clocks \"-o:$(generated_dir)/$(long_name).domains\""
```

PR #17, opened 2017-03-17
Fft changes

shunshou:
[DON'T MERGE]
Bug: If pads are used, clkgen transform will fail (can't figure out clk src through pad BB).


shunshou:
@colinschmidt I'm going to merge my changes -- local to my transforms, etc., so I can stay on master (to be compatible w/ craft stuff).

colinschmidt:
Go for it.

PR #22, opened 2017-04-04
utils.DiGraph was being being confused with the DigGraph in firrtl.

chick:
This led to pathological exceptions
No such method error on accessing a lazy val.
InstanceGraph seemed also to be a duplicate of firrtl code, and was used only in commented out code so I deleted it too.

***

IOPadSpec fails no two tests but these seem to be at least an ordinary error. And should be debugged separately


chick:
@stevobailey @shunshou @albert-magyar @colinschmidt 
This fixes a really bizarre method not found error/exception when running tests.  IOPadSpec fails to tests.  I don't think it's related to the DiGraph code but it might be.  Let me know if you think I need to dig into that

chick:
@colinschmidt Wow, that was a record push to merge.

shunshou:
Oh gosh sorry! Yeah it's b/c I stuck in Albert's stuff before it got merged with Firrtl. And I was running into the same issue when doing sbt publish-local so I commented out the use in ClkGen for now... I guess that wasn't the only problem. My bad!

shunshou:
I'm not using IOPad stuff for now (prob won't be in time for tapeout...), but I'll clean it up for a future tapeout hopefully around July.

PR #24, opened 2017-07-10
PCAD macro compiler port

donggyukim:
This is my effort not to use pcad. Of course, there's room for improvement, but I think this is good enough for the initial attempt.


colinschmidt:
Cool. Do you use this with some branch of PLSI or is it used in something different?

tapeout/src/test/resources/macros/lib-1024x8-mrw.json
```diff
@@ -0,0 +1,35 @@
...
+        "input port polarity": "active high",
+        "address port polarity": "active high",
+        "mask port polarity": "active high",
+        "chip enable port polarity": "active high",
+        "write enable port polarity": "active high"
+      }
+    ]
+  },
+  {
+    "type": "metal filler cell",
```
colinschmidt: Is this used anywhere?

tapeout/src/test/resources/macros/lib-2048x16-n28.json
```diff
@@ -0,0 +1,52 @@
+[
+  {
+    "type": "sram",
+    "name": "vendor_sram_16",
```
colinschmidt: Perhaps all the "name" fields should be consistent? I'm not sure what the benefit of having only some of them be different.

donggyukim:
Refer to ucb-bar/plsi#28.

tapeout/src/test/resources/macros/rocketchip.json
```diff
@@ -0,0 +1,76 @@
...
+        "address port name": "RW0_addr",
+        "mask port name": "RW0_wmask",
+        "chip enable port name": "RW0_en",
+        "write enable port name": "RW0_wmode"
+      }
+    ]
+  },
+  {
+    "type": "sram",
+    "name": "T_1090_ext",
```
colinschmidt: I'm a little confused about what this file represents? Why should it have chisel generated temporary names for the srams?

tapeout/src/test/resources/macros/saed32.json
```diff
@@ -0,0 +1,186 @@
+[
+  {
+    "type": "sram",
+    "name": "SRAM1RW1024x8",
```
colinschmidt: Is this technology public?

tapeout/src/test/resources/macros/lib-1024x8-mrw.json
donggyukim: Not for the macro compiler, but needed for floorplan, I guess. This also tests how well the macro compiler filters SRAMs only in JSON.

tapeout/src/test/resources/macros/lib-2048x16-n28.json
donggyukim: I copy and paste JSON files from pcad, only used for tests. Names are not important at all.

tapeout/src/test/resources/macros/rocketchip.json
donggyukim: This is the file generated from rocketchip, and I want to check if the macro compiler finds the correct mapping.

tapeout/src/test/resources/macros/saed32.json
donggyukim: I guess so, isn't it?

colinschmidt: A quick google shows its pretty exposed so I don't think this is an issue.

ben-k: The SAED stuff is not public - Synopsys keeps it locked up for use by its licensees only.  We've historically respected this for course materials (e.g., CS250) by keeping scripts behind the Berkeley wall.

That said, if it's already Google-compromised, then I guess I don't feel strongly about it.

colinschmidt:
Do you feel like the tests in here are good enough? I'm gonna go through the old C++ code and check for a valid translation I suppose.

tapeout/src/test/resources/macros/saed32.json
donggyukim: This file contains nothing special, so can we just rename it?

donggyukim:
@colinschmidt I'm not sure, but @edwardcwang wants to write more tests. It'll be great if you can validate the translation.

tapeout/src/main/scala/transforms/macros/Utils.scala
```diff
@@ -0,0 +1,154 @@
...
+import scala.language.implicitConversions
+
+trait PortPolarity
+case object ActiveLow extends PortPolarity
+case object ActiveHigh extends PortPolarity
+case object NegativeEdge extends PortPolarity
+case object PositiveEdge extends PortPolarity
+object PortPolarity {
+  implicit def toPortPolarity(s: Any): PortPolarity =
+    (s: @unchecked) match {
```
colinschmidt: Why not just check the missing cases and throw an error?

tapeout/src/main/scala/transforms/macros/Utils.scala
```diff
@@ -0,0 +1,154 @@
...
+    writeEnablePolarity: Option[PortPolarity],
+    maskName: Option[String],
+    maskPolarity: Option[PortPolarity],
+    maskGran: Option[BigInt],
+    width: BigInt,
+    depth: BigInt) {
+  val effectiveMaskGran = maskGran.getOrElse(width)
+  val AddrType = UIntType(IntWidth(ceilLog2(depth) max 1))
+  val DataType = UIntType(IntWidth(width))
+  val MaskType = UIntType(IntWidth(width / effectiveMaskGran))
```
colinschmidt: Naming for these should start with lowercase letters right?

tapeout/src/main/scala/transforms/macros/Utils.scala
```diff
@@ -0,0 +1,154 @@
...
+    (writeEnableName map (Field(_, Flip, BoolType))) ++
+    (maskName map (Field(_, Flip, MaskType)))
+  )
+  val ports = tpe.fields map (f => Port(
+     NoInfo, f.name, f.flip match { case Default => Output case Flip => Input }, f.tpe))
+}
+
+class Macro(lib: Map[String, Any]) {
+  val name = lib("name").asInstanceOf[String]
+  val width = BigInt(lib("width").asInstanceOf[Double].toLong)
```
colinschmidt: Why the intermediate conversion to Double?

tapeout/src/main/scala/transforms/macros/Utils.scala
```diff
@@ -0,0 +1,154 @@
...
+import java.io.File
+import scala.language.implicitConversions
+
+trait PortPolarity
+case object ActiveLow extends PortPolarity
+case object ActiveHigh extends PortPolarity
+case object NegativeEdge extends PortPolarity
+case object PositiveEdge extends PortPolarity
+object PortPolarity {
+  implicit def toPortPolarity(s: Any): PortPolarity =
```
colinschmidt: This has to be Any for the implicit to work? Do you ever call it with something other than a String?

tapeout/src/main/scala/transforms/macros/MacroCompiler.scala
```diff
@@ -0,0 +1,401 @@
...
+      if (i <= last + 1) {
+        /* Palmer: Every memory is going to have to fit at least a single bit. */
+        // coninue
+      } else if ((i - last) % lib.width.toInt == 0) {
+        /* Palmer: It's possible that we rolled over a memory's width here,
+                   if so generate one. */
+        pairs += ((last, i-1))
+        last = i
+      } else {
+        /* Palmer: FIXME: This is a mess, I must just be super confused. */
```
donggyukim: Yeah, everyone is confused here. We should fix it.

tapeout/src/main/scala/transforms/macros/MacroCompiler.scala
```diff
@@ -0,0 +1,401 @@
...
+            case (_, None) => // continue
+            case (_, Some(p)) if p == lib.width => // continue
+            case _ =>
+              System.err println "Bit-mask (or unmasked) target memories are suppored only"
+              return None
+          }
+        }
+      }
+    }
+    pairs += ((last, mem.width.toInt - 1))
```
donggyukim: `pairs` contains `(high, low)`s for bit extractions. It seems like bit extractions of PCAD IR must be `(high+1,low)`.

tapeout/src/main/scala/transforms/macros/MacroCompiler.scala
```diff
@@ -0,0 +1,401 @@
...
+          // Clock port mapping
+          /* Palmer: FIXME: I don't handle memories with read/write clocks yet. */
+          stmts += connectPorts(WRef(memPort.clockName),
+                                libPort.clockName,
+                                libPort.clockPolarity)
+
+          // Adress port mapping
+          /* Palmer: The address port to a memory is just the low-order bits of
+           * the top address. */
+          stmts += connectPorts(WRef(memPort.addressName),
```
donggyukim: That will be gracefully done by other passes.

tapeout/src/main/scala/transforms/macros/MacroCompiler.scala
```diff
@@ -0,0 +1,401 @@
...
+            System.err println s"INFO: unable to compile ${mem.name} using ${lib.name} port count must match"
+            (best, area)
+          case ((best, area), lib) =>
+            /* Palmer: A quick cost function (that must be kept in sync with
+             * memory_cost()) that attempts to avoid compiling unncessary
+             * memories.  This is a lower bound on the cost of compiling a
+             * memory: it assumes 100% bit-cell utilization when mapping. */
+            // val cost = 100 * (mem.depth * mem.width) / (lib.depth * lib.width) +
+            //                  (mem.depth * mem.width)
+            // Donggyu: I re-define cost
```
donggyukim: I didn't understand Palmer's cost model and couldn't translate it. I think my cost model is what it should do, considering the utilization and the number of macros.

tapeout/src/main/scala/transforms/macros/MacroCompiler.scala
```diff
@@ -0,0 +1,401 @@
...
+        best match {
+          case None => modules
+          case Some((mod, bb)) =>
+            (modules filterNot (m => m.name == mod.name || m.name == bb.name)) ++ Seq(mod, bb)
+        }
+      }
+      case _ => c.modules
+    }
+    val circuit = c.copy(modules = modules)
+    // print(circuit.serialize)
```
donggyukim: Sure, when it's ready to be merged.

tapeout/src/main/scala/transforms/macros/Utils.scala
donggyukim: Yeah, because JSON parser returns `Any`.

tapeout/src/main/scala/transforms/macros/Utils.scala
donggyukim: Well, it's essentially the same as it will throw mismatch exceptions. But I'm ok to change it.

tapeout/src/main/scala/transforms/macros/Utils.scala
donggyukim: Yeah, I just want to make them consistent for types. I'm also ok to change them.

tapeout/src/main/scala/transforms/macros/Utils.scala
donggyukim: Because the JSON parser always returns Double for numbers. There are errors with `Int` or `Long`.

donggyukim:
@colinschmidt @edwardcwang Yeah it's hard to validate. Unfortunately, it generates different Verilog from PCAD's. It's at least working with riscv-mini. I'll also try it with rocketchip soon. No hurries, we should have enough time to verify.

colinschmidt:
Can you show me a diff of the two verilog outputs?

donggyukim:
`sbt test` will give you the verilog outputs.

edwardcwang:
With regards to the JSON - I've documented the format/schema that PLSI/pcad uses to the best of my knowledge here: https://github.com/edwardcwang/plsi-mdf/blob/master/macro_format.yml

I'm planning on formatting the tests and all future development work off this standard, so I though it was about time to document this format (which Palmer did not document).

colinschmidt:
I chatted with @edwardcwang yesterday and gave him a bunch of sample inputs/outputs/confs from the hurricane-2 tapeout. Hopefully making sure these configs work will give us a good level of confidence in the compiler. My only worry is just the complexity of this code makes it a bit difficult to modify if we need to (i.e. I don't think I could do it even after the CR). Hopefully this is the last vlsi_mem_gen written at ucb-bar

donggyukim:
What do you mean by "complexity"? (style? algorithm?) I don't think this should be the final form of the compiler. The purpose of this PR is to "exactly" copy the functionality of pcad to firrtl. We should improve style & algorithm afterward.

colinschmidt:
Mostly style is my complaint with the current code. It seems difficult to make a large style change without what would amount to basically a rewrite IMO

donggyukim:
I totally agree. Let's focus on the functionality in this PR and improve it later.

shunshou:
I also started something similar when I was writing the repl_seq_mem pass. I'd really like to see split depth/width work with non power of 2 memory sizes (which is what I kind of got stuck on.../what I potentially need).

donggyukim:
@shunshou If you mean splitting non power of 2 memories using power of 2 memories, I think it does:
https://github.com/ucb-bar/barstools/blob/pcad/tapeout/src/test/scala/transforms/macros/SplitWidth.scala#L420
https://github.com/ucb-bar/barstools/blob/pcad/tapeout/src/test/scala/transforms/macros/SplitDepth.scala#L57

If you mean splitting memories using non power of 2 memories, I didn't try it yet.

shunshou:
If you want to split a memory with depth 6 (say the compiler only supports up to 5):
000
001
010
011
100
101

You've tried splitting it into a memory with depths 4 and 2,  so you can select via MSB. 

Or, if depths of 3 were legal (2 memories of depth 2), you can't strictly select by looking at MSB. You have to select as >= 3 and address the hight memory with original address - 3? 

Would you ever not want to split with only power of 2 memories?
If you had a depth of 1030, do you want to split it into 1024 + 8 ? Or 512 + 518?

donggyukim:
I don't think this resolves your issues. This compiler is very naive, so it only splits macros using a single block. We should also improve it to handle these cases, which is impossible if it's still in pcad...

shunshou:
What do you mean by using a single block? How does it split the 2000 depth memory? Does it make sure that all 2000 original addresses are properly mapped to unique memories/addresses?

colinschmidt:
I read single block as it will only pick one sram to map to and tile it appropriately. i.e. it would split 2000 -> 512*4 and never 2000-> 1024 + 512*2 or w/e

donggyukim:
So, `2000x8` can be split into two `1024x8` srams (1024 + 976) with MSB selecting an sram block, which may not be optimal though.

shunshou:
In general, is minimizing the # of SRAMs for a split better? There's additional macro placement, etc. overhead that's not really easy to account for.

aswaterman:
Splitting the RAMs into approximately-equal sizes will often be better for delay, but using the largest possible macros will often be better for area.  And then there's @shunshou's point about placement.  I don't think there's a one-size-fits-all heuristic...

colinschmidt:
I think the cost function should be user configurable. As it might change on a technology or project basis.

aswaterman:
Or even a more general user-specifiable callback (i.e. F(width, depth, list of available RAMs) -> list of chosen RAMs)

We can provide a few sample implementations of the callback, implementing some of the policies discussed here, like the euqalish-size heuristic.

shunshou:
I think I'd go for similar size + minimize # of splits (ease of placement, wire delay, etc.), and I think it's just a lot easier to reason about with power of 2 sizes, but would still like to understand tradeoffs w/ non power of 2.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "mdf"]
+	path = mdf
+	url = https://github.com/edwardcwang/plsi-mdf.git
```
donggyukim: @edwardcwang I didn't understand you mean a submodule by a library. We should avoid it.

edwardcwang: Let's discuss this in person then - I think in the interest of re-use, I'd like to have the macro description format be an independent library so I can use it for other passes and projects.

edwardcwang:
@donggyukim Concurrency issues should be fixed now.

donggyukim:
@edwardcwang a7577dd fixes a critical bug you should notice.

edwardcwang:
Good catch, thanks!

edwardcwang:
@donggyukim FYI, there is now a ConfReader in the mdf library.

edwardcwang:
@donggyukim I was working for the whole day today with a cleaner and more comprehensive implementation of the parallel compiler. Can you write a testcase for your bug so I can include it in my implementation? (which I think handles more cases than yours) My implementation is radically different and I don't think I can include your change as-is.

donggyukim:
Can you just delete mine and add your implementation? I only need to keep the history. I'll take a look then.

edwardcwang:
@donggyukim Yeah, I was planning to do that. I was asking for the testcase so that we can verify that my implementation doesn't introduce the bug you were trying to fix.

donggyukim:
Ok, I see your point. It appears when we want to map a single bit mask memories to non-masked libs whose width is larger than 1.

edwardcwang:
@donggyukim Okay, refactored parallel compiler is in. One thing: mapping a bitmasked mem to non-masked libs with width>1 can be extremely wasteful (since each lib is basically used to store a bit) - at that point you may just want to consider synflops.

edwardcwang:
@donggyukim @colinschmidt While it's not perfect, I think the memory compiler is now at the point where I would be happy with merging it. (And it supports byte-masked RAMs now for FPGA) Any last thoughts before I go ahead with the merge?

colinschmidt:
I'd like to get a quick CR in before its merged

edwardcwang:
@colinschmidt Do you want to do the quick CR then?

colinschmidt:
working on it. I'll give the go ahead at end of day with whatever partial CR I have by then

donggyukim:
@edwardcwang You shouldn't delete my bug fix commit or notify me. (I said I should keep history) I totally lost the pointer from other projects.

edwardcwang:
@donggyukim Sorry about that - I think I misunderstood what you meant in your comment. If you still have the revision ID we might still be able to find it in the git history.

donggyukim:
I have a temporal solution for now. In the future, we should be careful on sync.

edwardcwang:
I see. I think the current plan is to merge this mega PR as soon as I get to addressing Colin's comments, so I hope that the sync issues with this mega PR will disappear.

macros/src/main/scala/CostMetric.scala
```diff
@@ -0,0 +1,189 @@
...
+   *
+   * @param mem Memory macro to compile (target memory)
+   * @param lib Library memory macro to use (library memory)
+   * @return The cost of this compile, defined by this cost metric, or None if
+   *         it cannot be compiled.
+   */
+  def cost(mem: Macro, lib: Macro): Option[BigInt]
+
+  /**
+   * Helper function to return the map of argments (or an empty map if there are none).
```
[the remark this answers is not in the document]
edwardcwang: Got it, thanks.

macros/src/main/scala/CostMetric.scala
```diff
@@ -0,0 +1,189 @@
...
+
+  override def commandLineParams = Map()
+  override def name = "PalmerMetric"
+  override def construct(m: Map[String, String]) = PalmerMetric
+}
+
+/**
+ * An external cost function.
+ * Calls the specified path with paths to the JSON MDF representation of the mem
+ * and lib macros. The external executable should return a BigInt.
```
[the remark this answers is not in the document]
edwardcwang: Done

macros/src/main/scala/CostMetric.scala
```diff
@@ -0,0 +1,189 @@
...
+    pathOption match {
+      case Some(path:String) => new ExternalMetric(path)
+      case _ => throw new IllegalArgumentException("ExternalMetric missing option 'path'")
+    }
+  }
+}
+
+/** The current default metric in barstools, re-defined by Donggyu. */
+// TODO: write tests for this function to make sure it selects the right things
+object NewDefaultMetric extends CostMetric with CostMetricCompanion {
```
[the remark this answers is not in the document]
edwardcwang: Done.

macros/src/main/scala/MacroCompiler.scala
```diff
@@ -0,0 +1,635 @@
...
+  /** Macro compiler mode. */
+  sealed trait CompilerMode
+  /** Strict mode - must compile all memories or error out. */
+  case object Strict extends CompilerMode
+  /** Synflops mode - compile all memories with synflops (do not map to lib at all). */
+  case object Synflops extends CompilerMode
+  /** FallbackSynflops - compile all memories to SRAM when possible and fall back to synflops if a memory fails. **/
+  case object FallbackSynflops extends CompilerMode
+  /** Default mode - compile what is possible and do nothing with uncompiled memories. **/
+  case object Default extends CompilerMode
```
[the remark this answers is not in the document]
edwardcwang: Done. I left the default as CompileAvailable for now since FallbackSynflops isn't implemented yet and added a TODO.

macros/src/main/scala/MacroCompiler.scala
```diff
@@ -0,0 +1,635 @@
...
+   */
+  case class Params(mem: String, lib: Option[String], costMetric: CostMetric, mode: CompilerMode)
+
+  /**
+   * Create a MacroCompilerAnnotation.
+   * @param c Name of the module(?) for this annotation.
+   * @param p Parameters (see above).
+   */
+  def apply(c: String, p: Params): Annotation =
+    Annotation(CircuitName(c), classOf[MacroCompilerTransform], MacroCompilerUtil.objToString(p))
```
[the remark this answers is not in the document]
edwardcwang: It turns out that CircuitName is actually what we want after all. Added some more documentation in MacroCompilerAnnotation to make this clearer.

macros/src/main/scala/MacroCompiler.scala
```diff
@@ -0,0 +1,635 @@
...
+              // If the mem mask is a multiple of the lib width, then we're good.
+              // Just roll over every lib width as usual.
+              // e.g. lib width=4, mem maskGran={4, 8, 12, 16, ...}
+              splitMemory(libWidth)
+            } else if (libPort.src.width % p == 0) {
+              // Lib width is a multiple of the mem mask.
+              // Consider the case where mem mask = 4 but lib width = 8, unmasked.
+              // We can still compile, but will need to waste the extra bits.
+              splitMemory(memMask.get)
+            } else {
```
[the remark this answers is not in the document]
edwardcwang: I'm a bit hesitant to merge them directly since I'd like to guard against the `!(p < libPort.src.width)` case and I would prefer to keep them separate (at least I find it easier to read) as opposed to combining them into a else if (... & ...) clause above, though I'm happy to hear other thoughts on this.

macros/src/main/scala/MacroCompiler.scala
```diff
@@ -0,0 +1,635 @@
...
+              }
+            } else { // m < l
+              // Lib maskGran > mem maskGran.
+              if (l % m == 0) {
+                // Lib maskGran is a multiple of mem maskGran.
+                // e.g. lib maskGran = 8, mem maskGran = 4.
+                // In this case we can only compile very wastefully (by treating
+                // lib as a mem maskGran width memory) :(
+                splitMemory(memMask.get)
+              } else {
```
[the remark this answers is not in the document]
edwardcwang: In theory this is possible though I think the current scheme doesn't handle this (and it's much more complicated). I left a detailed comment with example.

macros/src/main/scala/MacroCompiler.scala
```diff
@@ -0,0 +1,635 @@
...
+              }
+            }
+          }
+        }
+      }
+    }
+    // Add in the last chunk if there are any leftovers
+    bitPairs += ((currentLSB, mem.src.width.toInt - 1))
+
+    // Serial mapping
```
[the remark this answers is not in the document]
edwardcwang: Sounds about right - renamed both.

macros/src/main/scala/MacroCompiler.scala
```diff
@@ -0,0 +1,635 @@
...
+            DoPrim(PrimOps.Eq, Seq(reg, index), Nil, index.tpe)
+        }
+        def andAddrMatch(e: Expression) = {
+          and(e, addrMatch)
+        }
+        val cats = ArrayBuffer[Expression]()
+        for (((low, high), j) <- bitPairs.zipWithIndex) {
+          val inst = WRef(s"mem_${i}_${j}", lib.tpe)
+
+          def connectPorts2(mem: Expression,
```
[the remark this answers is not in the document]
edwardcwang: Since these are "variables" (as opposed to methods), function overloading doesn't seem to work on them.

macros/src/main/scala/MacroCompiler.scala
```diff
@@ -0,0 +1,635 @@
...
+                                libPort.src.address.polarity)
+
+          // Output port mapping
+          (memPort.src.output, libPort.src.output) match {
+            case (Some(PolarizedPort(mem, _)), Some(PolarizedPort(lib, lib_polarity))) =>
+              /* Palmer: In order to produce the output of a memory we need to cat
+               * together a bunch of narrower memories, which can only be
+               * done after generating all the memories.  This saves up the
+               * output statements for later. */
+              val name = s"${mem}_${i}_${j}"
```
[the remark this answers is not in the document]
edwardcwang: I think the other WRef isn't the same (`val inst = WRef(s"mem_${i}_${j}", lib.tpe)` - `mem` vs `${mem}`). This WRef is the output from the particular instance, whereas the previous WRef is the IO reference for the entire instance.

macros/src/main/scala/MacroCompiler.scala
```diff
@@ -0,0 +1,635 @@
...
+                    bits(WRef(mem), (low + i) / memPort.src.effectiveMaskGran)
+                  }
+                })).reverse)
+              }
+            case None =>
+              /* Palmer: If there is no input port on the source memory port
+               * then we don't ever want to turn on this write
+               * enable.  Otherwise, we just _always_ turn on the
+               * write enable port on the inner memory. */
+              if (libPort.src.maskPort.isEmpty) one
```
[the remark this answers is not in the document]
edwardcwang: I think the convention seems to essentially treat a mask of width 1 to be a write enable, which is what the code below does. Not sure if it's the ultimately right thing to do (though it seems reasonable), but documented for now.

macros/src/main/scala/MacroCompiler.scala
```diff
@@ -0,0 +1,635 @@
...
+                }
+                case _ => // TODO: do we care about the case where mem has chipEnable but lib doesn't?
+              }
+            case (None, Some(PolarizedPort(we, we_polarity)), Some(PolarizedPort(en, en_polarity))) =>
+              // TODO
+              System.err.println("cannot emulate multi-bit mask ports with write enable")
+              return None
+            case (None, None, None) =>
+              /* Palmer: There's nothing to do here since there aren't any
+               * ports to match up. */
```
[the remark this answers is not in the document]
edwardcwang: This case is reached if it's a read-only port (in the case of memories with completely split read and write ports). Added a note.

macros/src/main/scala/MacroCompiler.scala
```diff
@@ -0,0 +1,635 @@
...
+               * write enable port instead of the mask port. */
+              stmts += connectPorts(andAddrMatch(and(memWriteEnable, memMask)),
+                                    we, we_polarity)
+              chipEnable match {
+                case Some(PolarizedPort(en, en_polarity)) => {
+                  stmts += connectPorts(andAddrMatch(memChipEnable), en, en_polarity)
+                }
+                case _ => // TODO: do we care about the case where mem has chipEnable but lib doesn't?
+              }
+            case (None, Some(PolarizedPort(we, we_polarity)), Some(PolarizedPort(en, en_polarity))) =>
```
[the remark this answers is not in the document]
edwardcwang: Done

colinschmidt:
Just go for it whenever you are ready

donggyukim:
I have few comments and concerns:
* When we merge it, we should squash it
* I have no idea how barstools is used by other people, but having mdf makes it difficult to submodule barstools/macro. Do we still want to keep mdf as a separate submodule? Is it still a bad idea to just use barstools/macro when someone needs mdf only?

edwardcwang:
I thought we discussed this already?

Also, I'd prefer to rebase to keep the history of the development.

donggyukim:
Yes, we discussed, but I just want to remind it again because macro compiler is hard to import. I'm not blocking it if everyone is fine with it.

Also, do you want to keep all 72 commits in master? Most of them are documentation and bump, so I think we should squash them at least.

edwardcwang:
My main thought for keeping all the commits in master is since we both put a lot of hard work into it, we may as well record our hard work in the git log.

edwardcwang:
I see. On the other hand, more commits is also an indicator that software is getting developed more actively and bugs are getting fixed :)

donggyukim:
@edwardcwang Just keep the history. I don't want you to waste your time.

edwardcwang:
Okay. Thanks for your understanding - hopefully going forward we have smaller PRs so we can avoid this type of problem.

PR #25, opened 2017-09-06
Added retime annotation

azidar:
They no longer compile with the latest Chisel/FIRRTL, and
may not be supported. However, future work will need them, so
this keeps the files around but are ignored by sbt.


colinschmidt:
Debating the merits of moving files versus deleting them. They will be in git history either way.

Retime annotation seems simple as it should be. Am I reading it correctly that instance annotations don't work?

shunshou:
I prefer moving vs. deleting. People forget they exist, otherwise. As an example: I had started work on the memory stuff way back when that got deleted... and then Donggyu and co. re-ported PCAD stuff over, creating potentially more redundant Scala work. 

I need some of the clkgen stuff (not the Firrtl part) for maintaining my code base, so I'd like not having to revive it entirely when I get back to that.

azidar:
So annotating an instance will create a Module annotation on the instance's module, instead of an annotation on the instance. This is ok if stuff isn't dedup'd, but I'd like Chisel to be able to create an instance annotation.

azidar:
So what's the decision, is .directory ok, or would you rather I comment out code (or delete it)? We can try this out and if its annoying to people, we can delete it later.

colinschmidt:
Sure you can leave the .directory.

The instance annotation is a limitation in general of annotations, or this particular one?

azidar:
Yeah, its a general limitation.

PR #27, opened 2017-10-06
Correct multi-ported memory compilation

### LINKED ISSUE #26: MacroCompiler is broken
At last, it's broken. It's not working with BOOM. I don't know which commit breaks it because there are a ton of rebased commits. In #24, I pointed out the problems which are not addressed.

The working version originally in the PR is:
https://github.com/ucb-bar/barstools/tree/donggyu-tmp

As we can see the history is totally different. I can't accept these changes. This is not about credits. The commits I pushed are **bug fixes**. They are all about functionalities. My commits are valuable enough to be preserved because I was directly applying it to rocket and boom, but they are disregarded. In #24, I was treated as if I have no knowledge of software engineering.

I'll point out again what's wrong with #24.
1.  The disaster starts from splitting mdf out from this repo. Now it's not clear the bug is in this repo or the mdf repo. On the other hand, mdf makes it much harder to import macro compiler. This repo is supposed to be imported to other projects, but I'm not sure why this is not disregarded. Obviously, this is a totally careless decision.
2. There are dozens of commits not related to functional changes.  I can't blame the commit among more than 50 commits.
3. The history should never be rewritten **in the same branch**. If you want to do, you should create another branch.
4. Refactoring should be done once everything is functionally working. In this sense, it's also very unfortunate the history is gone.
5. What exactly do the tests do? They don't help find a bug.

I just gave it up. I don't think I'm going to use barstools for midas/strober. Good luck to find a bug.

edwardcwang:
It was incorrectly splitting multiple times before. Fixed the issue and
added regression tests for this issue.

Close #26

### DIFF
--- a/macros/src/main/scala/MacroCompiler.scala
+++ b/macros/src/main/scala/MacroCompiler.scala
@@ -91,6 +91,17 @@ class MacroCompilerPass(mems: Option[Seq[Macro]],
                         libs: Option[Seq[Macro]],
                         costMetric: CostMetric = CostMetric.default,
                         mode: MacroCompilerAnnotation.CompilerMode = MacroCompilerAnnotation.Default) extends firrtl.passes.Pass {
+  // Helper function to check the legality of bitPairs.
+  // e.g. ((0,21), (22,43)) is legal
+  // ((0,21), (22,21)) is illegal and will throw an assert
+  private def checkBitPairs(bitPairs: Seq[(BigInt, BigInt)]): Unit = {
+    bitPairs.foldLeft(BigInt(-1))((lastBit, nextPair) => {
+      assert(lastBit + 1 == nextPair._1, s"Pair's first bit ${nextPair._1} does not follow last bit ${lastBit}");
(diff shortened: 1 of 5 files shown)


edwardcwang:
@donggyukim Could you try this fix to see if it addresses the BOOM issue? I did the regression test and discovered that the new parallel implementation did not account for multi-ported memories properly, so did a fix for that and wrote some tests.

donggyukim:
Ok, seems like it's working. I'll let you know when tests finished.

So, you're sure you take 1a2ad45953542a5ad506a614c3a6453d53e19242 into account carefully, right? This is very hard to catch when it's broken.

edwardcwang:
I believe so, though the rewrite is in a slightly different way which allows for non-bitmasked target memories. For extra assurance, if you gave me some cases which you think would have failed prior to 1a2ad45, I can run them against the current implementation and check the output.

donggyukim:
That appears with BOOM. So, if the outputs from BOOM config are the same, it would be fine.

donggyukim:
Good job for fixing the bug! Merge it whenever ready.

PR #28, opened 2017-12-03
Support firrtl output in command line for MacroCompiler

edwardcwang:
Fix a bug and support firrtl output

### DIFF
--- a/macros/src/main/scala/MacroCompiler.scala
+++ b/macros/src/main/scala/MacroCompiler.scala
@@ -611,6 +611,7 @@ object MacroCompiler extends App {
   case object Macros extends MacroParam
   case object Library extends MacroParam
   case object Verilog extends MacroParam
+  case object Firrtl extends MacroParam
   case object CostFunc extends MacroParam
   case object Mode extends MacroParam
   type MacroParamMap = Map[MacroParam, String]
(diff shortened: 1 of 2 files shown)


macros/src/main/scala/MacroCompiler.scala
```diff
@@ -677,18 +678,39 @@ object MacroCompiler extends App {
...
 
-      // Close the writer.
-      verilogWriter.close()
+            // Extract Verilog circuit and write it.
+            verilogWriter.write(result.getEmittedCircuit.value)
+
+            // Close the writer.
+            verilogWriter.close()
+          }
+          case None =>
```
colinschmidt: Since this isn't optional shouldn't you print an nice error message here? Or is this just going to fallback to printing the usage?

macros/src/test/scala/MacroCompilerSpec.scala
```diff
@@ -335,7 +335,7 @@ s"""
      * @param readEnable Has a read enable port?
      * @param mask Mask granularity (# bits) of the port or None. */
     def generateReadWriteFooterPort(prefix: String, readEnable: Boolean, mask: Option[Int]): String = {
-      generatePort(libPortPrefix, lib_addr_width, libWidth,
+      generatePort(prefix, lib_addr_width, libWidth,
```
colinschmidt: I think you should make the default argument for these prefixes the same as it was before.

edwardcwang: This specific change actually fixes a bug which prevented two of the unit tests from passing. I think the caller's arguments didn't change, so are there any other changes to make?

macros/src/main/scala/MacroCompiler.scala
edwardcwang: The idea is that you could choose firrtl output or Verilog output. Should I add a warning neither option is used?

colinschmidt: Oh I see. I don't think a neither check is necessary.

macros/src/test/scala/MacroCompilerSpec.scala
colinschmidt: Ah I didn't read far enough up in the diff. Yeah this is fine.

PR #29, opened 2018-02-14
Add memory compiler to macros

azidar:
Added memory compiler to mdf spec. You add the memory compiler to the lib file, and set the `--use-compiler` flag.

### DIFF
--- a/macros/src/main/scala/MacroCompiler.scala
+++ b/macros/src/main/scala/MacroCompiler.scala
@@ -68,7 +68,7 @@ object MacroCompilerAnnotation {
    * @param costMetric Cost metric to use
    * @param mode Compiler mode (see CompilerMode)
    */
-  case class Params(mem: String, lib: Option[String], costMetric: CostMetric, mode: CompilerMode)
+  case class Params(mem: String, lib: Option[String], costMetric: CostMetric, mode: CompilerMode, useCompiler: Boolean)
 
   /**
    * Create a MacroCompilerAnnotation.
@@ -121,15 +121,15 @@ class MacroCompilerPass(mems: Option[Seq[Macro]],
       for ((memPort, libPort) <- pairedPorts) {
 
         // Make sure we don't have a maskGran larger than the width of the memory.
-        assert (memPort.src.effectiveMaskGran <= memPort.src.width)
-        assert (libPort.src.effectiveMaskGran <= libPort.src.width)
+        assert (memPort.src.effectiveMaskGran <= memPort.src.width.get)
+        assert (libPort.src.effectiveMaskGran <= libPort.src.width.get)
 
-        val libWidth = libPort.src.width
+        val libWidth = libPort.src.width.get
 
         // Don't consider cases of maskGran == width as "masked" since those masks
         // effectively function as write-enable bits.
-        val memMask = if (memPort.src.effectiveMaskGran == memPort.src.width) None else memPort.src.maskGran
-        val libMask = if (libPort.src.effectiveMaskGran == libPort.src.width) None else libPort.src.maskGran
+        val memMask = if (memPort.src.effectiveMaskGran == memPort.src.width.get) None else memPort.src.maskGran
+        val libMask = if (libPort.src.effectiveMaskGran == libPort.src.width.get) None else libPort.src.maskGran
 
         (memMask, libMask) match {
           // Neither lib nor mem is masked.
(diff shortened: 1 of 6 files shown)


macros/src/test/scala/MacroCompilerSpec.scala
```diff
@@ -133,55 +138,69 @@ trait HasSRAMGenerator {
...
+  // Generate a "simple" SRAM group (active high/positive edge, 1 read-write port).
+  def generateSimpleSRAMGroup(prefix: String, mux: Int, depth: Range, width: Range, maskGran: Option[Int] = None, extraPorts: Seq[MacroExtraPort] = List()): SRAMGroup = {
+    SRAMGroup(Seq("mygroup_", "width", "x", "depth", "-", "VT"), "1rw", Seq("svt", "lvt", "ulvt"), mux, depth, width, Seq(generateReadWritePort(prefix, None, None, maskGran)))
+  }
+
+  // 'vt': ('svt','lvt','ulvt'), 'mux': 2,  'depth': range(16,513,8),       'width': range(8,289,2),   'ports': 1
+  // 'vt': ('svt','lvt','ulvt'), 'mux': 4,  'depth': range(32,1025,16),     'width': range(4,145),     'ports': 1}
+  def generateSRAMCompiler(name: String, prefix: String): mdf.macrolib.SRAMCompiler = {
+    SRAMCompiler(name, Seq(
+      generateSimpleSRAMGroup(prefix, 2, Range(16, 513, 8), Range(8, 289, 2)),
```
edwardcwang: 1) Are these meant to be hardcoded, or should this be parameterized?

2) Does it make more sense to have the ranges be inclusive (e.g. `[16, 512]`)? I had to think twice for a bit when seeing numbers like 513 and 1025.

macros/src/test/scala/SRAMCompiler.scala
```diff
@@ -0,0 +1,15 @@
+package barstools.macros
+
+import mdf.macrolib._
+
+class SRAMCompiler extends MacroCompilerSpec with HasSRAMGenerator {
```
edwardcwang: Can we have a test for the new facilities as well to show that it works? I can help with this if needed.

macros/src/test/scala/MacroCompilerSpec.scala
azidar: 1. Its just a helper function, so it can be parameterized however we want it to be, I don't care.
2. I took the numbers from stuff that Colin gave, but you're right I'm not sure whether the numbers are inclusive or not. We should check.

macros/src/test/scala/SRAMCompiler.scala
azidar: Yeah, I guess this doesn't check anything - it would be good to test it. I think it should be straightforward, but I'll take a crack at it later.

edwardcwang: Great. I just ask this since I caught so many bugs and edge cases by unit testing when I first wrote this code.

macros/src/main/scala/MacroCompiler.scala
```diff
@@ -121,15 +121,15 @@ class MacroCompilerPass(mems: Option[Seq[Macro]],
       for ((memPort, libPort) <- pairedPorts) {
 
         // Make sure we don't have a maskGran larger than the width of the memory.
-        assert (memPort.src.effectiveMaskGran <= memPort.src.width)
-        assert (libPort.src.effectiveMaskGran <= libPort.src.width)
+        assert (memPort.src.effectiveMaskGran <= memPort.src.width.get)
```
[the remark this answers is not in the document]
azidar: In this context, yes they will all be Some, because the ports are not on a SRAMCompiler. Its a bit of a hack, but then again, making width an Option[Int] does sort of make sense.

macros/src/main/scala/MacroCompiler.scala
```diff
@@ -68,7 +68,7 @@ object MacroCompilerAnnotation {
    * @param costMetric Cost metric to use
    * @param mode Compiler mode (see CompilerMode)
    */
-  case class Params(mem: String, lib: Option[String], costMetric: CostMetric, mode: CompilerMode)
+  case class Params(mem: String, lib: Option[String], costMetric: CostMetric, mode: CompilerMode, useCompiler: Boolean)
```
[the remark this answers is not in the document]
azidar: No, they are orthogonal. Perhaps there may be some future options, but we can cross that when we get there.

donggyukim:
This breaks midas, so can anyone explain what's going on?

edwardcwang:
Whoops, sorry about that. Could you describe the error/bug that emerges so that we can troubleshoot?

donggyukim:
I just got a bunch of compile errors. Seem like a lot of API changes...

edwardcwang:
In that case, should I try to reproduce the error using https://github.com/ucb-bar/midas-top or is there a better repo/branch given all the FireSim/MIDAS refactors?

donggyukim:
We can reproduce it in the dev branch of https://github.com/ucb-bar/midas-examples by bumping barstools

azidar:
My bad, just saw this. I'll take a look tomorrow and fix it.

edwardcwang:
@donggyukim Apologies for the delay. The fix for this is now in https://github.com/ucb-bar/midas/pull/73 - if we merge that and bump midas in midas-examples, the bug should be fixed.

PR #33, opened 2018-04-26
Add options to force certain memories to lib or synflops

edwardcwang:
Long-term plan is to add annotation support for this as well


azidar:
Sorry I'm just getting to this - is it still relevant?

edwardcwang:
Yes, this is still relevant and useful - e.g. when I'm taping out it's useful to be able to pin certain memories   that I know for sure should be flops.

PR #43, opened 2019-03-16
stop exceptions on empty conf files

abejgonzalez:
When running unit tests for `testchipip`, it errors since there are no memories generated in the `*.mems.conf`. Thus this PR is to address this by not excepting out.

### DIFF
--- a/macros/src/main/scala/MacroCompiler.scala
+++ b/macros/src/main/scala/MacroCompiler.scala
@@ -839,8 +839,7 @@ object MacroCompiler extends App {
           }
           case None =>
         }
-      }
-      else {
+      } else {
         // Warn user
         System.err println "WARNING: Empty *.mems.conf file. No memories generated."
 
--- a/macros/src/main/scala/MacroCompiler.scala
+++ b/macros/src/main/scala/MacroCompiler.scala
@@ -840,6 +840,20 @@ object MacroCompiler extends App {
           case None =>
         }
       }
+      else {
+        // Warn user
+        System.err println "WARNING: Empty *.mems.conf file. No memories generated."
+
+        // Emit empty verilog file if no macros found
+        params.get(Verilog) match {
+          case Some(verilogFile: String) => {
+            // Create an empty verilog file
+            val verilogWriter = new FileWriter(new File(verilogFile))
+            verilogWriter.close()
+          }
+          case None =>
+        }
+      }
     } catch {
       case e: java.util.NoSuchElementException =>
         e.printStackTrace()
--- a/macros/src/main/scala/MemConf.scala
+++ b/macros/src/main/scala/MemConf.scala
@@ -49,8 +49,7 @@ object MemConf {
   def fromString(s: String): Seq[MemConf] = {
     if (s.isEmpty) {
       Seq[MemConf]()
-    }
-    else {
+    } else {
       s.split("\n").toSeq.map(_ match {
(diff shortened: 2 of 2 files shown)


abejgonzalez:
Should this also be mentioned/fixed in `firrtl`? Should there be no `*.mems.conf` emitted in this case?

colinschmidt:
I think you should issue a warning in this case. I think most people who run MacroCompiler expect it to have memories to compile so we should notify them we aren't going to do anything.

edwardcwang:
One minor caveat: MacroCompiler could be run as part of an automated process/flow where the parameters could be set in a way such that no memories are generated. Maybe have a flag `--allow-empty` or something to accommodate these usecases?

abejgonzalez:
Another issue that may come up is that if the `*.mems.conf` file is not found, and this fix is put in, then there will be no `*.mems.v` file emitted which can cause VCS compile to fail.

In this case, it might be good to emit an empty `.v` file for VCS to consume?

As for @edwardcwang's point, I think having this flag is fine... although, I think that anyone who has an automated flow for building/running sims (that uses macro compiler as the last step in the process) will most likely have the flag enabled by default (since BAR's normal unit tests most likely generate these mem stubs). It actually think that it would be more helpful for people running it outside of this flow.

jwright6323:
IMO the default behavior should be to emit an empty *.mems.conf (with a warning, perhaps one that could be squelched with `--no-warn-if-empty`) for the build system reasons already stated. I don't think that exiting/erroring is useful. If you are getting to the point where you are running MacroCompiler and are unsure if you have SRAMs you have bigger problems.

jwright6323:
Also, yes I think you should emit an empty *.mems.v to make the build system happy.

colinschmidt:
The current behavior from firrtl is good. Empty conf file when no memories. Emitting an empty verilog file already happens sometimes in MacroCompiler so just adding another case seems fine.
I think a single line of warning will prevent unwanted emails/issue later when someone says "MacroCompiler" runs but produced an empty file.

colinschmidt:
Does sbt test still pass?

abejgonzalez:
`sbt test` passes all test except for the `BOOMTest` in the top-level directory. However, this doesn't look like an issue with my changes.

colinschmidt:
I'm pretty sure I just fixed that can you rebase on master?

abejgonzalez:
Ok with the rebase everything works!

macros/src/main/scala/MacroCompiler.scala
```diff
@@ -840,6 +840,20 @@ object MacroCompiler extends App {
           case None =>
         }
       }
+      else {
```
colinschmidt: Can we put this on the same line as closing bracket above

macros/src/main/scala/MemConf.scala
```diff
@@ -47,9 +47,14 @@ object MemConf {
...
 
   def fromString(s: String): Seq[MemConf] = {
-    s.split("\n").toSeq.map(_ match {
-      case MemConf.regex(name, depth, width, ports, maskGran) => MemConf(name, depth.toInt, width.toInt, MemPort.fromString(ports), Option(maskGran).map(_.toInt))
-      case _ => throw new Exception(s"Error parsing MemConf string : ${s}")
-    })
+    if (s.isEmpty) {
+      Seq[MemConf]()
+    }
+    else {
```
colinschmidt: Same style comment as above.

colinschmidt:
Will push the button once travis is done.

PR #45, opened 2019-03-17
SRAM mapping improvements

colinschmidt:
This is a multi-feature PR with each commit being a small feature.

Chronologically the features are:

- Allow SRAMs to have ports without associated clocks. This is the typical kind of SRAM encountered in the wild. Ideally we would put some compile-time assertion that the chisel memories read and write clocks are the same but I'm leaving this as a TODO for now. We simply connect one of the clocks to the SRAM clock port.
- Next we fix the default cost model. Cost is changed to a double from BigInt to avoid funny integer off by one junk. Then we drop the last term that was causing the inverse of expected cost. In rocket-chip designs with a commercial SRAM compiler this now picks the largest possible memory whereas the previous version of this metric picked the smallest possible memories.
- Then we improve MacroCompilers handling of generative SRAM compilers. We filter the SRAMs by number of ports and mask compatibility and then by ability to provide integral mappings for the proposed design memory. This results in a large selection but was kept in for fancy cost metrics to utilize. The runtime is a few seconds with all the integral mappings and is not noticeably longer than only the largest mapping.
- Finally we change the input forms of the passes used by GenerateTop such that we replace the SeqMems after we have removed the TestHarness. This avoids creating a conf with very large memories that would break downstream tools, such as MacroCompiler.

### DIFF
--- a/macros/src/main/scala/CostMetric.scala
+++ b/macros/src/main/scala/CostMetric.scala
@@ -18,7 +18,7 @@ trait CostMetric extends Serializable {
    * @return The cost of this compile, defined by this cost metric, or None if
    *         it cannot be compiled.
    */
-  def cost(mem: Macro, lib: Macro): Option[BigInt]
+  def cost(mem: Macro, lib: Macro): Option[Double]
 
   /**
    * Helper function to return the map of arguments (or an empty map if there are none).
@@ -43,7 +43,7 @@ trait CostMetricCompanion {
  * TODO: figure out what is the difference between this metric and the current
  * default metric and either revive or delete this metric. */
 object OldMetric extends CostMetric with CostMetricCompanion {
-  override def cost(mem: Macro, lib: Macro): Option[BigInt] = {
+  override def cost(mem: Macro, lib: Macro): Option[Double] = {
     /* Palmer: A quick cost function (that must be kept in sync with
      * memory_cost()) that attempts to avoid compiling unncessary
      * memories.  This is a lower bound on the cost of compiling a
@@ -61,9 +61,9 @@ object OldMetric extends CostMetric with CostMetricCompanion {
 /**
  * An external cost function.
  * Calls the specified path with paths to the JSON MDF representation of the mem
- * and lib macros. The external executable should print a BigInt.
+ * and lib macros. The external executable should print a Double.
  * None will be returned if the external executable does not print a valid
- * BigInt.
+ * Double.
  */
 class ExternalMetric(path: String) extends CostMetric {
   import mdf.macrolib.Utils.writeMacroToPath
@@ -71,7 +71,7 @@ class ExternalMetric(path: String) extends CostMetric {
   import scala.language.postfixOps // for !! postfix op
   import sys.process._
 
-  override def cost(mem: Macro, lib: Macro): Option[BigInt] = {
+  override def cost(mem: Macro, lib: Macro): Option[Double] = {
     // Create temporary files.
     val memFile = File.createTempFile("_macrocompiler_mem_", ".json")
     val libFile = File.createTempFile("_macrocompiler_lib_", ".json")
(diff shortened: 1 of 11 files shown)


macros/src/main/scala/CostMetric.scala
```diff
@@ -113,18 +113,16 @@ object ExternalMetric extends CostMetricCompanion {
 /** The current default metric in barstools, re-defined by Donggyu. */
 // TODO: write tests for this function to make sure it selects the right things
 object DefaultMetric extends CostMetric with CostMetricCompanion {
-  override def cost(mem: Macro, lib: Macro): Option[BigInt] = {
+  override def cost(mem: Macro, lib: Macro): Option[Double] = {
     val memMask = mem.src.ports map (_.maskGran) find (_.isDefined) map (_.get)
     val libMask = lib.src.ports map (_.maskGran) find (_.isDefined) map (_.get)
     val memWidth = (memMask, libMask) match {
-      case (Some(1), Some(1)) | (None, _) => mem.src.width
+      case (Some(_), Some(1)) | (None, _) => mem.src.width
```
jwright6323: Was this just a bug? Not sure I understand the change.

macros/src/main/scala/CostMetric.scala
```diff
@@ -113,18 +113,16 @@ object ExternalMetric extends CostMetricCompanion {
...
       case (Some(p), _) => p // assume that the memory consists of smaller chunks
     }
-    return Some(
-      (((mem.src.depth - 1) / lib.src.depth) + 1) *
-      (((memWidth - 1) / lib.src.width) + 1) *
-      (lib.src.depth * lib.src.width + 1) // weights on # cells
-    )
+    val depthCost = (mem.src.depth.toDouble / lib.src.depth.toDouble)
+    val widthCost = (memWidth.toDouble / lib.src.width.toDouble)
+    return Some(depthCost * widthCost)
```
jwright6323: I know that we currently don't support using larger library SRAM macros than the requested ones, but this cost function would break if we did. Maybe add an assertion or TODO for now?

macros/src/main/scala/MacroCompiler.scala
```diff
@@ -561,9 +564,32 @@ class MacroCompilerPass(mems: Option[Seq[Macro]],
         // in the 'circuit'.
         (mems foldLeft c.modules){ (modules, mem) =>
 
+        val sram = mem.src
+        def groupMatchesMask(group: SRAMGroup, mem:SRAMMacro): Boolean = {
+          val memMask = mem.ports map (_.maskGran) find (_.isDefined) map (_.get)
+          val libMask = group.ports map (_.maskGran) find (_.isDefined) map (_.get)
+          (memMask, libMask) match {
+            case (_, Some(1)) => true
```
jwright6323: Isn't this a subset of cases 2 and 4 below?

macros/src/main/scala/CostMetric.scala
colinschmidt: If the library has a mask granularity of 1 we can fit any source/design memory in it so we can just consider its width the nominal width (rather than some function of its mask width and the library mask width).
This function is here for some edge that someone was trying to work around but in practice never shows up. This code is slightly more correct but the second case I think is still not correct.

macros/src/main/scala/CostMetric.scala
colinschmidt: Thats a good point. We actually do support larger library macros if we don't have a compiler.
I actually broke the old function for that use case. Will think about this a bit more.

macros/src/main/scala/MacroCompiler.scala
colinschmidt: Good catch thanks!

colinschmidt:
@edwardcwang I've fixed the unit tests that were extant (broken before I started). I think thats enough for me for now.

PR #48, opened 2019-03-18
Add Travis

colinschmidt:
Work in progress


colinschmidt:
It passes! Let us never break the tests again.
@edwardcwang @jwright6323

PR #49, opened 2019-03-18
HighForm has whens so we need to check for instances there

colinschmidt:
Fixes a bug


colinschmidt:
@azidar Is this the only IR construct that contains statements thats added when going from Mid->High form?

PR #53, opened 2019-03-27
Fix annos

### LINKED ISSUE #36: More intelligently filter annotations during reparenting
After #35 The `ReParentCircuit` transform removes modules but doesn't smartly filter annotations (instead it just removes all `DontTouchAnnotation`s). Filtering on module name didn't seem to work for me.

colinschmidt:
Fixes #36 and other problems related to renaming annotations correctly

### DIFF
--- a/tapeout/src/main/scala/transforms/RemoveUnusedModules.scala
+++ b/tapeout/src/main/scala/transforms/RemoveUnusedModules.scala
@@ -5,7 +5,7 @@ package barstools.tapeout.transforms
 import firrtl._
 import firrtl.ir._
 import firrtl.passes.Pass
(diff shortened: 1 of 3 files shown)


tapeout/src/main/scala/transforms/RemoveUnusedModules.scala
```diff
@@ -53,17 +53,9 @@ class RemoveUnusedModules extends Transform {
     val renames = state.renames.getOrElse(RenameMap())
 
     // This is what the annotation filter should look like, but for some reason it doesn't work.
```
jwright6323: This comment can die now :)

jwright6323:
LGTM, @azidar should probably take a look too

PR #54, opened 2019-03-29
Add options to emit top/harness firrtl and annotations

jwright6323:
This adds options to emit the firrtl for the top and harness, as well as the re-pathed annotations.

### DIFF
--- a/tapeout/src/main/scala/transforms/Generate.scala
+++ b/tapeout/src/main/scala/transforms/Generate.scala
@@ -37,6 +37,28 @@ trait HasTapeoutOptions { self: ExecutionOptionsManager with HasFirrtlOptions =>
       "use this to set synTop"
     }
 
+  parser.opt[String]("top-fir")
+    .abbr("tsf")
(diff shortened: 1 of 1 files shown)


tapeout/src/main/scala/transforms/Generate.scala
```diff
@@ -99,7 +147,26 @@ sealed trait GenerateTopAndHarnessApp extends LazyLogging { this: App =>
       customTransforms = firrtlOptions.customTransforms ++ topTransforms
     )
 
-    firrtl.Driver.execute(optionsManager)
+    val result = firrtl.Driver.execute(optionsManager)
+
+    result match {
```
colinschmidt: Do you care about what level of firrtl is emitted? This will be the lowest level but you could also do the emission after the last custom transform to get a higher level of firrtl.

jwright6323: I think the lowest level is probably fine here, unless you see a reason to produce a higher form.

PR #57, opened 2019-05-06
Support Bigint mems

colinschmidt:
This also collapses synflops generation to be closer to normal macro generation. Synflops no longer does memory splitting itself instead providing a simple memory for the normal macro compiler to map to.

This breaks CI currently because there is no snapshot of firrtl-interpreter with BigInt mems.


edwardcwang:
Generally seems fine to me though I'd be curious to understand why synflops needed to split memories in the first place?

colinschmidt:
It was an entirely separate path so it would make Vec based memories to avoid dealing with masking.

PR #63, opened 2019-07-18
Coordinate Top and Harness generation

albert-magyar:
This avoids name conflicts between the design top and the test harness by coordinating the two FIRRTL compiler runs into one invocation of the main generator app. It also makes Verilog simulator generation somewhat faster.

A few tweaks allow the two runs to perform their independent tasks while avoiding namespace collisions:
1.) A new `AvoidExtModuleCollisions` pass appends the `ExtModules` that appear in the top circuit to the harness circuit, which prevents their names from being used by passes that create new modules (like repl-seq-mem).
2.) The options get changed slightly from the first top FIRRTL run to the second harness run (output filename changes, conf file location changes, custom transform list changes).
3.) A command line flag gets added to specify the location of the conf file for the harness memories.
4.) The blackbox resource annotations from the top circuit get passed to the harness circuit state, which results in a single, unified `firrtl_black_box_resource_files.f` file being created.

Chipyard PR: ucb-bar/chipyard#168

### DIFF
--- a/tapeout/src/main/scala/transforms/Generate.scala
+++ b/tapeout/src/main/scala/transforms/Generate.scala
@@ -172,22 +172,21 @@ sealed trait GenerateTopAndHarnessApp extends LazyLogging { this: App =>
     annotations = firrtlOptions.annotations ++ tapeoutOptions.topDotfOut.map(BlackBoxResourceFileNameAnno(_))
   )
 
-  class AvoidExtModuleCollisions(mustLink: CircuitState) extends Transform {
+  class AvoidExtModuleCollisions(mustLink: Seq[ExtModule]) extends Transform {
     def inputForm = HighForm
     def outputForm = HighForm
-    val extModules = mustLink.circuit.modules.collect{ case e: ExtModule => e }
     def execute(state: CircuitState): CircuitState = {
-      state.copy(circuit = state.circuit.copy(modules = state.circuit.modules ++ extModules))
+      state.copy(circuit = state.circuit.copy(modules = state.circuit.modules ++ mustLink))
     }
   }
 
-  private def harnessTransforms(linkedTop: CircuitState): Seq[Transform] = {
+  private def harnessTransforms(topExtModules: Seq[ExtModule]): Seq[Transform] = {
     // XXX this is a hack, we really should be checking the masters to see if they are ExtModules
     val externals = Set(harnessTop.get, synTop.get, "SimSerial", "SimDTM")
     Seq(
       new ConvertToExtMod((m) => m.name == synTop.get),
       new RemoveUnusedModules,
-      new AvoidExtModuleCollisions(linkedTop),
+      new AvoidExtModuleCollisions(topExtModules),
       new RenameModulesAndInstances((old) => if (externals contains old) old else (old + "_in" + harnessTop.get))
     )
   }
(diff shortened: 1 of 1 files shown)


tapeout/src/main/scala/transforms/Generate.scala
```diff
@@ -129,81 +143,99 @@ sealed trait GenerateTopAndHarnessApp extends LazyLogging { this: App =>
     )
   }
 
+  class AvoidExtModuleCollisions(mustLink: CircuitState) extends Transform {
+    def inputForm = HighForm
+    def outputForm = HighForm
+    val extModules = mustLink.circuit.modules.collect{ case e: ExtModule => e }
+    def execute(state: CircuitState): CircuitState = {
+      state.copy(circuit = state.circuit.copy(modules = state.circuit.modules ++ extModules))
```
jwright6323: Is this where the renaming happens?

albert-magyar:
It does do that — is that a problem?

jwright6323:
Maybe... ongoing discussion here: https://github.com/ucb-bar/chipyard/pull/168/files/30729cf6a105217a4505286a839010ac4cce27db#r305563772

jwright6323:
@albert-magyar How hard would it be to emit a `.f` file for just the top module? Now that they're uniquified it should be easy to then remove those from the combined `.f` to create the harness `.f`

albert-magyar:
@jwright6323 I think there are two fairly simple ways to do that:
- Patch FIRRTL to make it possible to supply a non-hardcoded name for the resources `.f` file. I wrote the (fairly minimal) code to do this, but I am not sure whether it would get upstreamed.
- Supply different target-dir arguments (each a subdirectory of the current target-dir) for the top and harness and deal with their disparate locations in the Makefile.

jwright6323:
@albert-magyar IMO the first bullet would be useful even beyond this PR. Do you want to do it or do you want someone else to help out?

albert-magyar:
@jwright6323 I opened a firrtl PR (freechipsproject/firrtl:1129) for option #1.

jwright6323:
Cool, thanks!

alonamid:
Is this still waiting on something?

abejgonzalez:
This LGTM once the CI passes.

albert-magyar:
It won’t pass since the chipyard one isn’t in :)

I can just merge this one now, though?

abejgonzalez:
Isn't the CI just a `sbt test` and it shouldn't depend on Chipyard being merged in?

abejgonzalez:
As for merging in... I think it is tested enough (with the Chipyard CI) so I think this is safe to merge. Though I think @jwright6323 @colinschmidt probably need to give the final say.

albert-magyar:
The files won’t compile since it’s using an older version of FIRRTL. I’m not sure how the buildbot is publishing FIRRTL in its local environment.

abejgonzalez:
It looks like the updated FIRRTL has not been pushed to the Sonatype repo. So I think this is fine. I personally don't know how you would go around this other than maybe doing a clone of Chipyard, building the FIRRTL jar, then adding it to the classpath so it doesn't pull from Sonatype.

albert-magyar:
That's what I had assumed it was doing (publishing from a clone of
Chipyard).

PR #66, opened 2019-08-17
Remove large annotations

abejgonzalez:
This PR removes large annotations that were deleted but kept inside of the output `json`. This fixes an OOM error on `beagle`, but I suspect it will happen on other very large designs.

Do we need any of these annotations for other tooling that happens after the `Generate...` steps?

### DIFF
--- a/tapeout/src/main/scala/transforms/Generate.scala
+++ b/tapeout/src/main/scala/transforms/Generate.scala
@@ -202,13 +202,7 @@ sealed trait GenerateTopAndHarnessApp extends LazyLogging { this: App =>
     annoFile.foreach { annoPath =>
       val outputFile = new java.io.PrintWriter(annoPath)
       outputFile.write(JsonProtocol.serialize(res.circuitState.annotations.filter(_ match {
-        case DeletedAnnotation(_, anno) =>
-          anno match {
-            case ec: EmittedComponent => false
-            case ea: EmittedAnnotation[_] => false
-            case fca: FirrtlCircuitAnnotation => false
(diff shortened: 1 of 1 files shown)


seldridge:
I *think* you only need these annotations if you are using FIRRTL or Chisel Drivers. Those need to construct execution results and that logic has to kludgily rely on deleted annotations. If you are fully on ChiselStage and FirrtlStage it should be fine to delete these.

It's been a while, but I thought there were some options to either remove these automatically or optionally remove these. 

Working from memory here, so this info may be wrong.

albert-magyar:
@seldridge, I think this is just filtering them for the manual dumping of annotations to a JSON file that happens in the barstools flow. Would just dropping all `DeletedAnnotation`s be okay?

seldridge:
@albert-magyar: dropping all `DeletedAnnotation`s from the output JSON is totally fine. In fact, all `Stage`s will do this automatically for you (unless you disable this behavior with `WriteDeletedAnnotation`/`--write-deleted`). Ignore my comments above...

abejgonzalez:
Thanks for the clarification @seldridge and for the PR fix @albert-magyar! This LGTM. @colinschmidt @jwright6323 do we care about travis passing?

colinschmidt:
Yes we care about travis passing generally.

abejgonzalez:
@colinschmidt @jwright6323 Is this good to go? If so, Ill go ahead and merge.

jwright6323:
@abejgonzalez we should fix CI in a separate PR, though

PR #68, opened 2019-10-24
Print the firrtl exception if we get one

### LINKED ISSUE #67: Print exceptions when firrtl compilation fails
Don't just say illegal firrtl when firrtl has a nice error message see Generate.scala:222

colinschmidt:
Fixes #67

### DIFF
--- a/tapeout/src/main/scala/transforms/Generate.scala
+++ b/tapeout/src/main/scala/transforms/Generate.scala
@@ -218,7 +218,7 @@ sealed trait GenerateTopAndHarnessApp extends LazyLogging { this: App =>
       case x: FirrtlExecutionSuccess =>
         dump(x, tapeoutOptions.topFir, tapeoutOptions.topAnnoOut)
         x.circuitState.circuit.modules.collect{ case e: ExtModule => e }
-      case e =>
+      case x =>
         throw new Exception(s"executeTop failed while executing FIRRTL!\n${e}")
     }
   }
@@ -243,7 +243,7 @@ sealed trait GenerateTopAndHarnessApp extends LazyLogging { this: App =>
     val harnessResult = firrtl.Driver.execute(optionsManager)
     harnessResult match {
       case x: FirrtlExecutionSuccess => dump(x, tapeoutOptions.harnessFir, tapeoutOptions.harnessAnnoOut)
-      case e => throw new Exception(s"executeHarness failed while executing FIRRTL!\n${e}")
+      case x => throw new Exception(s"executeHarness failed while executing FIRRTL!\n${e}")
     }
   }
 }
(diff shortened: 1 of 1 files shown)


tapeout/src/main/scala/transforms/Generate.scala
```diff
@@ -218,8 +218,8 @@ sealed trait GenerateTopAndHarnessApp extends LazyLogging { this: App =>
       case x: FirrtlExecutionSuccess =>
         dump(x, tapeoutOptions.topFir, tapeoutOptions.topAnnoOut)
         x.circuitState.circuit.modules.collect{ case e: ExtModule => e }
-      case _ =>
-        throw new Exception("executeTop failed on illegal FIRRTL input!")
+      case e =>
```
jwright6323: just a nit, but we should probably use the same case variable (`e` vs `x`) for both cases for clarity- one `e` is an `ExtModule` and the other is a result

tapeout/src/main/scala/transforms/Generate.scala
```diff
@@ -243,7 +243,7 @@ sealed trait GenerateTopAndHarnessApp extends LazyLogging { this: App =>
     val harnessResult = firrtl.Driver.execute(optionsManager)
     harnessResult match {
       case x: FirrtlExecutionSuccess => dump(x, tapeoutOptions.harnessFir, tapeoutOptions.harnessAnnoOut)
-      case _ =>
+      case e => throw new Exception(s"executeHarness failed while executing FIRRTL!\n${e}")
```
jwright6323: same comment although this is less bad

colinschmidt:
Fixed reviewer feedback

tapeout/src/main/scala/transforms/Generate.scala
```diff
@@ -218,8 +218,8 @@ sealed trait GenerateTopAndHarnessApp extends LazyLogging { this: App =>
       case x: FirrtlExecutionSuccess =>
         dump(x, tapeoutOptions.topFir, tapeoutOptions.topAnnoOut)
         x.circuitState.circuit.modules.collect{ case e: ExtModule => e }
-      case _ =>
-        throw new Exception("executeTop failed on illegal FIRRTL input!")
+      case x =>
+        throw new Exception(s"executeTop failed while executing FIRRTL!\n${e}")
```
abejgonzalez:
```suggestion
        throw new Exception(s"executeTop failed while executing FIRRTL!\n${x}")
```

tapeout/src/main/scala/transforms/Generate.scala
```diff
@@ -243,7 +243,7 @@ sealed trait GenerateTopAndHarnessApp extends LazyLogging { this: App =>
     val harnessResult = firrtl.Driver.execute(optionsManager)
     harnessResult match {
       case x: FirrtlExecutionSuccess => dump(x, tapeoutOptions.harnessFir, tapeoutOptions.harnessAnnoOut)
-      case _ =>
+      case x => throw new Exception(s"executeHarness failed while executing FIRRTL!\n${e}")
```
abejgonzalez:
```suggestion
      case x => throw new Exception(s"executeHarness failed while executing FIRRTL!\n${x}")
```

PR #70, opened 2019-11-05
Fix MacroCompiler for CE-less Library Memories

abejgonzalez:
If a memory doesn't have a mask and doesn't have a chip enable, make sure that you use the `mem` chip enable to connect to the `we` port on the `lib` memory. Fixes a bug where the `lib` `we` signal would be tied to the `mem` `wmode` signal but then the macro would have no `en` signal connected to it.


abejgonzalez:
I will add one soon, I marked you all as reviewers to see if there was any pushback on the changes. IMO the more invasive one is the `re` change.

PR #73, opened 2019-12-10
Rc bump aug 2019

colinschmidt:
Updates for rocket-chip and chisel 3.2.0 for chipyard see: ucb-bar/chipyard#358


colinschmidt:
Thanks.
Blame past me for CI failure I think. (there may be an intermediate person to blame but I can't remember who)

PR #78, opened 2020-03-17
Add IO cell models

jwright6323:
This is a work in progress (this doesn't even elaborate yet), but basically this is my goal for how we can connect the IO cells in the ChipTop chipyard layer while keeping most of the IO types consistent with their System counterparts. Obviously excluded in that objective are widgets like GPIOs which have multiple control signals that we want to drive the IO cells with.

### DIFF
--- a/iocell/src/main/scala/chisel/IOCell.scala
+++ b/iocell/src/main/scala/chisel/IOCell.scala
@@ -6,11 +6,31 @@ import chisel3._
 import chisel3.util.{Cat, HasBlackBoxResource}
 import chisel3.experimental.{Analog, DataMirror, IO}
 
+// The following four IO cell bundle types are bare-minimum functional connections
+// for modeling 4 different IO cell scenarios. The intention is that the user
+// would create wrapper modules that extend these interfaces with additional
+// control signals. These are loosely similar to the sifive-blocks PinCtrl bundles
+// (https://github.com/sifive/sifive-blocks/blob/master/src/main/scala/devices/pinctrl/PinCtrl.scala),
+// but we want to avoid a dependency on an external libraries.
+
+/**
+ * The base IO bundle for an analog signal (typically something with no digital buffers inside)
+ *  pad: off-chip (external) connection
+ *  core: internal connection
+ */
 class AnalogIOCellBundle extends Bundle {
-  val pad = Analog(1.W)
-  val core = Analog(1.W)
+  val pad = Analog(1.W)   // Pad/bump signal (off-chip)
+  val core = Analog(1.W)  // core signal (on-chip)
 }
 
+/**
+ * The base IO bundle for a signal with runtime-controllable direction
+ *  pad: off-chip (external) connection
+ *  i: input to chip logic (output from IO cell)
+ *  ie: enable signal for i
+ *  o: output from chip logic (input to IO cell)
+ *  oe: enable signal for o
+ */
 class DigitalGPIOCellBundle extends Bundle {
   val pad = Analog(1.W)
   val i = Output(Bool())
@@ -19,19 +39,31 @@ class DigitalGPIOCellBundle extends Bundle {
   val oe = Input(Bool())
 }
 
+/**
+ * The base IO bundle for a digital output signal
+ *  pad: off-chip (external) connection
+ *  o: output from chip logic (input to IO cell)
+ *  oe: enable signal for o
+ */
 class DigitalOutIOCellBundle extends Bundle {
   val pad = Output(Bool())
   val o = Input(Bool())
   val oe = Input(Bool())
 }
 
+/**
+ * The base IO bundle for a digital input signal
+ *  pad: off-chip (external) connection
+ *  i: input to chip logic (output from IO cell)
+ *  ie: enable signal for i
+ */
 class DigitalInIOCellBundle extends Bundle {
   val pad = Input(Bool())
   val i = Output(Bool())
   val ie = Input(Bool())
 }
 
-abstract class IOCell extends BlackBox with HasBlackBoxResource
+abstract class IOCell extends BlackBox
 
 abstract class AnalogIOCell extends IOCell {
   val io: AnalogIOCellBundle
@@ -49,24 +81,28 @@ abstract class DigitalOutIOCell extends IOCell {
   val io: DigitalOutIOCellBundle
 }
 
-class GenericAnalogIOCell extends AnalogIOCell {
+// The following Generic IO cell black boxes have verilog models that mimic a very simple
+// implementation of an IO cell. For building a real chip, it is important to implement
+// and use similar classes which wrap the foundry-specific IO cells.
+
+trait GenericIOCell extends HasBlackBoxResource {
+  addResource("/barstools/iocell/vsrc/IOCell.v")
+}
+
+class GenericAnalogIOCell extends AnalogIOCell with IsGenericIOCell {
   val io = IO(new AnalogIOCellBundle)
-  addResource("/barstools/iocell/vsrc/IOCell.v")
 }
 
-class GenericDigitalGPIOCell extends DigitalGPIOCell {
+class GenericDigitalGPIOCell extends DigitalGPIOCell with IsGenericIOCell {
   val io = IO(new DigitalGPIOCellBundle)
-  addResource("/barstools/iocell/vsrc/IOCell.v")
 }
 
-class GenericDigitalInIOCell extends DigitalInIOCell {
+class GenericDigitalInIOCell extends DigitalInIOCell with IsGenericIOCell {
   val io = IO(new DigitalInIOCellBundle)
-  addResource("/barstools/iocell/vsrc/IOCell.v")
 }
 
-class GenericDigitalOutIOCell extends DigitalOutIOCell {
+class GenericDigitalOutIOCell extends DigitalOutIOCell with IsGenericIOCell {
   val io = IO(new DigitalOutIOCellBundle)
-  addResource("/barstools/iocell/vsrc/IOCell.v")
 }
 
 object IOCell {
@@ -76,6 +112,16 @@ object IOCell {
   def genericInput() = Module(new GenericDigitalInIOCell)
   def genericOutput() = Module(new GenericDigitalOutIOCell)
 
+  /**
+   * From within a RawModule or MultiIOModule context, generate new module IOs from a given
+   * signal and return the new IO and a Seq containing all generated IO cells.
+   * @param coreSignal The signal onto which to add IO cells
+   * @param name An optional name or name prefix to use for naming IO cells
+   * @param inFn A function to generate a DigitalInIOCell to use for input signals
+   * @param outFn A function to generate a DigitalOutIOCell to use for output signals
+   * @param anaFn A function to generate an AnalogIOCell to use for analog signals
+   * @return A tuple of (the generated IO data node, a Seq of all generated IO cell instances)
+   */
   def generateIOFromSignal[T <: Data](coreSignal: T, name: Option[String] = None,
     inFn: () => DigitalInIOCell = IOCell.genericInput,
     outFn: () => DigitalOutIOCell = IOCell.genericOutput,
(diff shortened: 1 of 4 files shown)


iocell/src/main/scala/chisel/IOCell.scala
```diff
@@ -0,0 +1,188 @@
...
+
+abstract class DigitalInIOCell extends IOCell {
+  val io: DigitalInIOCellBundle
+}
+
+abstract class DigitalOutIOCell extends IOCell {
+  val io: DigitalOutIOCellBundle
+}
+
+class ExampleAnalogIOCell extends AnalogIOCell {
```
colinschmidt: I'm not sure I like the `Example` prefix. They are more like models of an IO cell, or an RTL only IO cell.

iocell/src/main/scala/chisel/IOCell.scala
```diff
@@ -0,0 +1,188 @@
...
+  val io = IO(new AnalogIOCellBundle)
+  addResource("/barstools/iocell/vsrc/IOCell.v")
```
colinschmidt: Should we just move this into the abstract class?

iocell/src/main/scala/chisel/IOCell.scala
```diff
@@ -0,0 +1,188 @@
...
+}
+
+class ExampleDigitalOutIOCell extends DigitalOutIOCell {
+  val io = IO(new DigitalOutIOCellBundle)
+  addResource("/barstools/iocell/vsrc/IOCell.v")
+}
+
+object IOCell {
+
+  def exampleAnalog() = Module(new ExampleAnalogIOCell)
```
colinschmidt: Are these wrappers really doing anything?

iocell/src/main/scala/chisel/IOCell.scala
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  def generateFromSignal[T <: Data](coreSignal: T, padSignal: T, name: Option[String] = None,
+    inFn: () => DigitalInIOCell = IOCell.exampleInput,
+    outFn: () => DigitalOutIOCell = IOCell.exampleOutput,
+    anaFn: () => AnalogIOCell = IOCell.exampleAnalog): Seq[IOCell] =
+  {
+    coreSignal match {
+      case coreSignal: Analog => {
+        if (coreSignal.getWidth == 0) {
```
colinschmidt: Is it possible to hoist this compare out of all the individual cases?

iocell/src/main/scala/chisel/IOCell.scala
```diff
@@ -0,0 +1,188 @@
...
+          }
+          case ActualDirection.Output => {
+            val iocell = outFn()
+            name.foreach(n => iocell.suggestName(n))
+            iocell.io.o := coreSignal.asUInt.asBool
+            iocell.io.oe := true.B
+            padSignal := iocell.io.pad.asClock
+            Seq(iocell)
+          }
+          case _ => throw new Exception("Unknown direction")
```
colinschmidt: I think this exception should be more descriptive.
"Clock IO does not have direction and cannot be matched to an IOCell"
or something like that.

iocell/src/main/scala/chisel/IOCell.scala
```diff
@@ -0,0 +1,188 @@
...
+            case _ => {}
+          }
+          Seq()
+        } else {
+          DataMirror.directionOf(coreSignal) match {
+            case ActualDirection.Input => {
+              // this type cast is safe because we guarantee that padSignal and coreSignal are the same type (T), but the compiler is not smart enough to know that
+              val iocells = padSignal.asInstanceOf[Bits].asBools.zipWithIndex.map { case (w, i) =>
+                val iocell = inFn()
+                name.foreach(n => iocell.suggestName(n + "_" + i))
```
colinschmidt: It might be worth letting chisel do the deduplication of overlapping names here. That way it would match the normal chisel naming conventions.

iocell/src/main/scala/chisel/IOCell.scala
```diff
@@ -0,0 +1,188 @@
...
+                iocell.io.pad := w
+                iocell.io.ie := true.B
+                iocell
+              }
+              coreSignal := Cat(iocells.map(_.io.i).reverse)
```
colinschmidt: Should definitely have a comment about the `reverse` here, maybe also point to where the similar thing is done in chisel/firrtl.

iocell/src/main/scala/chisel/IOCell.scala
```diff
@@ -0,0 +1,188 @@
...
+              iocells
+            }
+            case ActualDirection.Output => {
+              val iocells = coreSignal.asBools.zipWithIndex.map { case (w, i) =>
```
colinschmidt: Why are is the letter `w`? I would have assumed it would be something like `c` for core? or even `core`?

iocell/src/main/scala/chisel/IOCell.scala
```diff
@@ -0,0 +1,188 @@
...
+              }
+              padSignal := Cat(iocells.map(_.io.pad).reverse)
+              iocells
+            }
+            case _ => throw new Exception("Unknown direction")
+          }
+        }
+      }
+      case coreSignal: Vec[Data] => {
+        // this type cast is safe because we guarantee that padSignal and coreSignal are the same type (T), but the compiler is not smart enough to know that
```
colinschmidt: I wonder if there is a way to match on both in the case to avoid the cast?

iocell/src/main/scala/chisel/IOCell.scala
```diff
@@ -0,0 +1,188 @@
...
+import chisel3._
+import chisel3.util.{Cat, HasBlackBoxResource}
+import chisel3.experimental.{Analog, DataMirror, IO}
+
+class AnalogIOCellBundle extends Bundle {
+  val pad = Analog(1.W)
+  val core = Analog(1.W)
+}
+
+class DigitalGPIOCellBundle extends Bundle {
```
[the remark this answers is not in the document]
jwright6323: They are more or less the same idea; I just didn't want to add a sifive-blocks dependency to barstools. I'll add a comment that says such.

iocell/src/main/scala/chisel/IOCell.scala
jwright6323: Yeah, not super attached to it myself. Is `Generic` better?

iocell/src/main/scala/chisel/IOCell.scala
jwright6323: No, because I want users to be able to extend the abstract class with their own implementations, which would have a different resource. I thought about making an `ExampleIOCell` trait (can name it something else) that does this but it's about the same amount of characters, so it doesn't buy much.

iocell/src/main/scala/chisel/IOCell.scala
jwright6323: Not sure what you mean. The example* functions? It mostly just makes passing these functions as arguments a little cleaner.

iocell/src/main/scala/chisel/IOCell.scala
jwright6323: No- the comparison is handled differently depending on the underlying type. For `Analog`, specifically, widths of 0 or > 1 are impossible to support, so we have to assert. For most other things, any widths are fine.

iocell/src/main/scala/chisel/IOCell.scala
jwright6323: Sure, will fix.

iocell/src/main/scala/chisel/IOCell.scala
jwright6323: It can't; you get anonymous names.

iocell/src/main/scala/chisel/IOCell.scala
jwright6323: OK, I'll add a comment. This is mostly just a bad Chisel API, IMO, but I understand "Why"

iocell/src/main/scala/chisel/IOCell.scala
jwright6323: Uhh I think it was w for wire? Honestly, I went through a few iterations of the code without changing the loop variable names, so it doesn't really make sense anymore. I'll fix it.

iocell/src/main/scala/chisel/IOCell.scala
jwright6323: I can try matching on a tuple of both. Let me play with that.

iocell/src/main/scala/chisel/IOCell.scala
colinschmidt: Sure `Generic` is better.

iocell/src/main/scala/chisel/IOCell.scala
colinschmidt: Hmm...ok. I think we should discuss extension use cases more concretely (and write documentation about how they would work).

iocell/src/main/scala/chisel/IOCell.scala
colinschmidt: Ok.

iocell/src/main/scala/chisel/IOCell.scala
colinschmidt: But is returning an empty Seq acceptable for all zero width signals?

iocell/src/main/scala/chisel/IOCell.scala
colinschmidt: I meant `iocell.suggestName(n)`

iocell/src/main/scala/chisel/IOCell.scala
jwright6323: Unfortunately no, you have to handle the `Bits` case specially (scroll down to bits to see the `0.U` dummy assignment)

iocell/src/main/scala/chisel/IOCell.scala
jwright6323: I'd be worried that it would not number them in the same order

colinschmidt: Ok I'm worried that we will have things like `myIO_0, myIO_1` whereas the core signal is `myIO, myIO_1`

jwright6323: I see. hmm. Let's discuss offline. I'd blame Chisel for this as well :).

iocell/src/main/scala/chisel/IOCell.scala
jwright6323: Fixed

iocell/src/main/scala/chisel/IOCell.scala
jwright6323: Cool, the tuple worked.

iocell/src/main/scala/chisel/IOCell.scala
jwright6323: This is fixed

iocell/src/main/scala/chisel/IOCell.scala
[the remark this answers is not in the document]
jwright6323: Done

iocell/src/main/scala/chisel/IOCell.scala
jwright6323: I added some explanation inline... if it's insufficient let's chat more.

iocell/src/main/scala/chisel/IOCell.scala
jwright6323: Done

iocell/src/main/scala/chisel/IOCell.scala
jwright6323: Fixed

iocell/src/main/scala/chisel/IOCell.scala
jwright6323: Changed this as requested

jwright6323:
@colinschmidt verbally approved this, so merging

PR #79, opened 2020-04-15
Update IOCell generation to handle abstract and async reset

davidbiancolin:
I need to fix the scaladoc by moving around the concretization functions around but otherwise this is ready for review.

This for the RC bump (ucb-bar/chipyard#503).

### DIFF
--- a/iocell/src/main/scala/chisel/IOCell.scala
+++ b/iocell/src/main/scala/chisel/IOCell.scala
@@ -120,15 +120,19 @@ object IOCell {
    * @param inFn A function to generate a DigitalInIOCell to use for input signals
    * @param outFn A function to generate a DigitalOutIOCell to use for output signals
    * @param anaFn A function to generate an AnalogIOCell to use for analog signals
+   * @param abstractResetAsAsync When set, will coerce abstract resets to
+   *        AsyncReset, and otherwise to Bool (sync reset)
    * @return A tuple of (the generated IO data node, a Seq of all generated IO cell instances)
    */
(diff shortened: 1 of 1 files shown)


iocell/src/main/scala/chisel/IOCell.scala
```diff
@@ -143,12 +147,47 @@ object IOCell {
...
+  val toAsyncReset: (Reset) => AsyncReset = _.asAsyncReset
+  def generateFromSignal[T <: Data, R <: Reset](
+    coreSignal: T,
+    padSignal: T,
+    name: Option[String] = None,
     inFn: () => DigitalInIOCell = IOCell.genericInput,
     outFn: () => DigitalOutIOCell = IOCell.genericOutput,
-    anaFn: () => AnalogIOCell = IOCell.genericAnalog): Seq[IOCell] =
+    anaFn: () => AnalogIOCell = IOCell.genericAnalog,
+    concretizeResetFn : (Reset) => R = toSyncReset): Seq[IOCell] =
```
jwright6323, line 159: Is there a reason to do this here rather than mirror the `abstractResetAsAsync` argument from  `generateIOFromSignal`? Just curious why the arguments are slightly different.

davidbiancolin: I wrote this code first and i wanted to mirror the argument list above, before realizing that user probably only cares about two cases. I'll make it consistent in using the bool. It simplifies the type signature of the function too.

PR #80, opened 2020-05-06
Firrtl 1.3 bump

albert-magyar:


macros/src/main/scala/MacroCompiler.scala
```diff
@@ -834,36 +834,24 @@ object MacroCompiler extends App {
...
           ))
         )
-        // Append a NoDCEAnnotation to avoid dead code elimination removing the non-parent SRAMs
-        val state = CircuitState(circuit, HighForm, annotations :+ NoDCEAnnotation)
-
-        // Run the compiler.
-        val result = new MacroCompiler().compileAndEmit(state)
-
-        // Write output FIRRTL file.
-        params.get(Firrtl) match {
```
colinschmidt, line 844: Can we file an issue about this removal to discuss/keep track of who needed it.

macros/src/main/scala/MacroCompiler.scala
```diff
@@ -834,36 +834,24 @@ object MacroCompiler extends App {
...
-          case None =>
-        }
 
-        // Write output Verilog file.
-        params.get(Verilog) match {
-          case Some(verilogFile: String) => {
-            // Open the writer for the output Verilog file.
-            val verilogWriter = new FileWriter(new File(verilogFile))
+        // The actual MacroCompilerTransform basically just generates an input circuit
+        val macroCompilerInput = CircuitState(circuit, HighForm, annotations)
```
colinschmidt, line 839: I would have assumed this form would match the `getLoweringTransforms` origin form from before.

tapeout/src/main/scala/transforms/AddSuffixToModuleNames.scala
```diff
@@ -0,0 +1,65 @@
...
+case class ModuleNameSuffixAnnotation(target: CircuitTarget, suffix: String)
+    extends SingleTargetAnnotation[CircuitTarget] {
+  def duplicate(n: CircuitTarget) = this.copy(target = n)
+}
+
+// This doesn't rename ExtModules under the assumption that they're some
+// Verilog black box and therefore can't be renamed.  Since the point is to
+// allow FIRRTL to be linked together using "cat" and ExtModules don't get
+// emitted, this should be safe.
+class AddSuffixToModuleNames extends Transform {
```
colinschmidt, line 25: Thanks for fixing this ancient kludge.

macros/src/main/scala/MacroCompiler.scala
albert-magyar: It's actually matching the `CircuitState` input form from before.

However, in reality, it should actually match the `inputForm` of `MacroCompilerTransform`, which is `midForm`, so I'll change that.

albert-magyar, line 839:
```suggestion
        val macroCompilerInput = CircuitState(circuit, MidForm, annotations)
```

macros/src/main/scala/MacroCompiler.scala
albert-magyar: Good idea, I put that in #81

macros/src/main/scala/MacroCompiler.scala
albert-magyar: Before and after this PR, the flow goes Input -> MacroCompilerTransform -> LoweringTransforms. This state I mark the CircuitState with is only applied for the duration of the MacroCompilerTransform, since it gets serialized after, so it just has to agree with that one transform.

PR #83, opened 2020-06-04
Getting barstools to support building the FFT generator

chick:
Bump versions of chisel3 and chisel-iotesters
Bump scalaVersion to 2.12.10
Bump sbt version to 1.3.8
Do a bunch of icky stuff so IsClkModule works
Change a bunch of Vec calls to VecInit
Change several chiselCloneType to chiselTypeOf

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,14 +1,14 @@
 // See LICENSE for license details.
 
 val defaultVersions = Map(
-  "chisel3" -> "3.2-SNAPSHOT",
-  "chisel-iotesters" -> "1.3-SNAPSHOT"
+  "chisel3" -> "3.3-SNAPSHOT",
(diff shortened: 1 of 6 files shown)


chick:
I'm not sure why any of this was necessary, perhaps I'm missing something in my build environment, but I gave up trying to figure it out, because it was blocking any progress on the FFT. Hopefully somebody can point out to me what's going on here.

edwardcwang:
Are you referring to dsptools FFT? or some other FFT?

chick:
Angie's FFT, it's in a private repo

colinschmidt:
I would recommend using the versions of chisel et al that are used in chipyard I believe that is 3.3.x and 1.3.x, as that will have the most robust support with CI running and such.

chick:
Closing this for now.

PR #86, opened 2020-09-05
Clean up IOCell types and parameterization

jerryz123:


colinschmidt (approved):
LGTM

iocell/src/main/scala/chisel/IOCell.scala
```diff
@@ -63,76 +63,68 @@ class DigitalInIOCellBundle extends Bundle {
   val ie = Input(Bool())
 }
 
-abstract class IOCell extends BlackBox
+trait IOCell extends BaseModule
 
-abstract class AnalogIOCell extends IOCell {
-  val io: AnalogIOCellBundle
```
jwright6323, line 69: I don't think you should have changed this; it was done this way intentionally to allow the user to extend the IOCellBundle with foundry IO cell pins. Can you have multiple `IO`s in `BlackBox`es now?

iocell/src/main/scala/chisel/IOCell.scala
```diff
@@ -63,76 +63,68 @@ class DigitalInIOCellBundle extends Bundle {
   val ie = Input(Bool())
 }
 
-abstract class IOCell extends BlackBox
+trait IOCell extends BaseModule
```
jwright6323, line 66: Why does this need to be a `BaseModule`? The only way to drive `Analog` pins is a `BlackBox`.

jwright6323 (commented):
A little late since this was merged already, but I think this actually regresses on some features and should be partially reverted.

iocell/src/main/scala/chisel/IOCell.scala
jerryz123: FireSim's IOCell models cannot be blackboxes

iocell/src/main/scala/chisel/IOCell.scala
jerryz123: I can add this feature back in

jwright6323: Cool, thanks. I think everything else looks good, for the record.

jerryz123: Actually, it seems like you can do this in the current system: 
```
class FancyDigitalGPIOCellBundle extends DigitalGPIOCellBundle {
  val extra = Output(Bool())
}

class FancyDigitalGPIOCell extends GenericIOCell with DigitalGPIOCell {
  override val io = IO(new FancyDigitalGPIOCellBundle)
}
```
so no changes are necessary?

jwright6323: I don't think `override` works- you'll call `IO` twice and end up with phantom `IO`s that are disconnected (also I don't think you can call `IO` twice in `BlackBox` anyway)

jerryz123: NVM the above doesn't work.

PR #87, opened 2020-09-14
Move to Chisel 3.4, Firrtl 1.4, remove .clk and .pad stuff

chick:
Refactor tapeout for Chisel 3.4, Firrtl 1.4

- Remove clk package based on discussion with Colin
- Annotations need to be refactored to using latest API
  - Generally that involves making annos generated by a anonymous ChiselAnnotation
  - The chisel annotations will use RunFirrtlTransform to queue up its associated transform
  - Chisel annotation provieds toFirrtl to generate Firrtl form of annotation
- Usages of unapply on firrtl annotations cannot use generic unapply(target, transform, data) which has been eliminated
- Have transforms use with DependencyAPIMigration to avoid deprecated `form`s
- Added some 'see License comments
- TechnologyLocation section of AddIOPadsSpec does not currently run because there is no content for it.
  - Added some tests for annotation serialization here
- FoundryPadsYaml would not parse json
  - Made separate case class for data
  - Now parses
  - Fails later with UnknownType in firrt compiler


tapeout/src/main/scala/transforms/AddSuffixToModuleNames.scala
```diff
@@ -22,9 +21,7 @@ case class ModuleNameSuffixAnnotation(target: CircuitTarget, suffix: String)
 // Verilog black box and therefore can't be renamed.  Since the point is to
 // allow FIRRTL to be linked together using "cat" and ExtModules don't get
 // emitted, this should be safe.
-class AddSuffixToModuleNames extends Transform {
-  def inputForm = LowForm
-  def outputForm = LowForm
+class AddSuffixToModuleNames extends Transform with DependencyAPIMigration {
```
colinschmidt, line 24: How is the ordering of these transforms handled now? I thought we were supposed to add which transforms it needs to run after (i.e. MidToLow or something).

colinschmidt (commented):
I'm concerned that without the `LowForm` ordering requirements we are losing information that will result in bad pass ordering.
I also think we should create a chipyard PR that bumps barstools to this PR to see what else breaks. I don't think the unit tests in here even smoke test all the functionality used in chipyard.

chick:
@colinschmidt Sorry for the delay, release stuff got in the way
I have now added the transform ordering logic.
Most of the tests run except for two of the AddIOPads.
I'm not to sure what's going on there, but since it, like the .clk stuff, has the hiding period I wasn't sure if
maybe it can go away as well. Please advies

colinschmidt (commented):
I think its fine to also remove the pads library.
I think we need to test this in situ from chipyard as well.

chick:
Pads Library stuff has been removed.
Can you give me a pointer to how I might test this in ChipYard.

colinschmidt:
I'm not sure if it will be easy but perhaps if you start the branch/PR others can chip-in.
I'd make a branch off of chipyard/dev, then bump barstools to this branch, and rocket-chip to its 3.4/1.4 RC branch, and try to build a simulator by typing `make` in `sims/verilator`.
Presumably some of the generators integrated in chipyard will have compile issues but I think if you start the branch/PR it is slightly easier for the maintainers of those generators to see the issues and fix them.

timsnyder-siv:
Hey @chick , I pushed the changes I needed in MacroCompiler.scala for it to work with FIRRTL 1.4 to https://github.com/sifive/barstools/commit/fc3a3eabff12b28dac417027dc3da62f9838876f  Would it make sense for me to open a PR to have it merged into the branch for this PR and then we can start the Chipyard PR?  I already started working on an updated Chipyard & Firesim branch that I'd be happy to push.  It probably makes sense for the barstools changes to be merged into a common branch to get that started.

timsnyder-siv:
I created https://github.com/ucb-bar/barstools/pull/92 to get my changes merged in here.

chick:
@timsnyder-siv I'm still just feeling my way around chipyard so I am not savvy on how they do branch management, so I am open to suggestions but that that sounds like a good idea.

colinschmidt:
@timsnyder-siv I think you can setup the chipyard branch and have it point to your commit. If you want to merge it into this branch first that is probably ok, since we'll be able to change it after the chipyard PR is created anyways.
I think top priority is to make the chipyard branch you have public so people can see all the changes that are needed outside of firesim&barstools.

timsnyder:
I created the chipyard PR earlier this week.  It is https://github.com/ucb-bar/chipyard/pull/717.  I just updated the pointer in that PR to point back here instead of at my PR.

tapeout/src/main/scala/transforms/AddSuffixToModuleNames.scala
```diff
@@ -18,30 +19,36 @@ case class ModuleNameSuffixAnnotation(target: CircuitTarget, suffix: String)
...
 
+  def determineRenamerandAnnos(state: CircuitState): (AnnotationSeq, (String) => String) = {
+    // remove determine suffix annotation
+    val newAnnos = state.annotations.filterNot(_.isInstanceOf[ModuleNameSuffixAnnotation])
+    val suffixes = state.annotations.collect({ case ModuleNameSuffixAnnotation(_, suffix) => suffix })
+    require(suffixes.length <= 1)
     val suffix = suffixes.headOption.getOrElse("")
-    val renamer = { name: String => if (whitelist(name)) name else name + suffix }
+
+    // skip renaming ExtModules and top-level module
```
abejgonzalez, line 36: One notable change is that this transform automatically skips renaming `ExtModule`s and the `state.circuit.main` module. This allows you to not have a list of `externals` that shouldn't be renamed in the `Generate.scala` file.

colinschmidt:
We should make sure to include the MDF scala bump and update this scala version as well,
see: https://github.com/ucb-bar/plsi-mdf/pull/6

colinschmidt:
It seems like we need to update the CI (hopefully just removing the dead tests from the code we deleted)
```
[info] *** 4 TESTS FAILED ***
[error] Failed: Total 4, Failed 4, Errors 0, Passed 0
[error] Failed tests:
[error] 	barstools.tapeout.transforms.retime.test.RetimeSpec
[error] 	barstools.tapeout.transforms.ResetNSpec
```

colinschmidt (approved):
LGTM

PR #88, opened 2020-09-14
Undo regression in iocell flexibility

jerryz123:


jwright6323 (approved):
LGTM, thanks for fixing

PR #92, opened 2020-10-22
Update MacroCompiler for Chisel 3.4 / FIRRTL 1.4

timsnyder-siv:
FireSim needs MacroCompiler.scala and these are the changes @jackkoenig and I made to get it working with FIRRTL 1.4.


colinschmidt (approved):
This seems relatively rote. LGTM

macros/src/main/scala/MacroCompiler.scala
```diff
@@ -92,16 +99,9 @@ object MacroCompilerAnnotation {
    * @param c Top-level circuit name (see class description)
    * @param p Parameters (see above).
    */
-  def apply(c: String, p: Params): Annotation =
-    Annotation(CircuitName(c), classOf[MacroCompilerTransform], MacroCompilerUtil.objToString(p))
+  def apply(c: String, p: Params): MacroCompilerAnnotation =
+    MacroCompilerAnnotation(MacroCompilerUtil.objToString(p))
```
jackkoenig, line 103: If anyone is wondering why the factory method pattern is useful, this is why. We were able to change the implementation of this method and it's actually binary compatible (not that binary compatibility really matters in this case).

macros/src/main/scala/MacroCompiler.scala
```diff
@@ -656,9 +656,9 @@ class MacroCompilerTransform extends Transform {
   def inputForm = MidForm
   def outputForm = MidForm
 
-  def execute(state: CircuitState) = getMyAnnotations(state) match {
-    case Seq(MacroCompilerAnnotation(state.circuit.main,
-    MacroCompilerAnnotation.Params(memFile, memFileFormat, libFile, hammerIR, costMetric, mode, useCompiler, forceCompile, forceSynflops))) =>
+  def execute(state: CircuitState) = state.annotations match {
+    case Seq(anno: MacroCompilerAnnotation) =>
```
jackkoenig, line 660: Wait, is this right? Shouldn't we like collect the `MacroCompilerAnnotation` first?

colinschmidt: Did `getMyAnnotations` do the collect before?

jackkoenig: Yeah, collect the ones that mark this transform

timsnyder-siv:
I'm realizing that I only pushed our changes to MacroCompiler but it would be good to be able to pass `test` in sbt project macros.  Wondering if that would have caught the issue regarding the need to collect the annotations in `barstools.macros.MacroCompilerTransform#execute`.  I'm updating the build.sbt on my branch and cleaning up Chisel3.4 stuff in the tests for macros too.

colinschmidt:
There is also a travis failure that is potentially related.

timsnyder-siv:
Updating build.sbt and runinng `project macros; test` in sbt shows that `barstools.macros.GenerateSomeVerilog` is failing because the verilog generated is empty.  It took me awhile looking at the testing code to understand that this is probably the collection thing @jackkoenig noticed and already mentioned.  I pushed my update to build.sbt and the macros tests but now it seems like there are conflicts with the firrtl-1.4-remove-clk-stuff branch.  I'm going to merge that branch into mine and repush.

macros/src/main/scala/MacroCompiler.scala
timsnyder-siv: Provided you meant doing:
```
def execute(state: CircuitState) = state.annotations.collect { case a: MacroCompilerAnnotation => a } match {
```

I resolved this in 446cb84

timsnyder-siv:
Hmmm.  Not sure why I have a couple of tests failing in the `tapeout` namespace failing after merging the `firrtl-1.4-remove-clk-stuff` branch into this PR.  I'll try to reproduce that locally.  It was my first time using the github web UI to merge and resolve conflicts.  Perhaps I pooched something.

PR #95, opened 2021-02-04
Make the directory structure match the packages

chick:
All tests run as they did prior to the changes


abejgonzalez:
Generally, I don't have a problem with this... though up to this point none of the other Chipyard projects do this. We mainly do something like this: `package <REPO_NAME>.<SBT_PROJECT>.<SOMETHING_1>.<SOMETHING_2>` and the `src/main` path would be something like `<REPO_NAME>/<SBT_PROJECT>/src/main/scala/<SOMETHING_1>/<SOMETHING_2>`. For example, this would be the following for something in `barstools`:

```
package barstools.tapeout.transforms.retime

// gives

barstools/tapeout/src/main/scala/transforms/retime
```

Is there a reason why you followed this method?

Also, it looks like the `utils` under the `src/main/scale/tapeout/transforms/utils` has the wrong package name... shouldn't it be `tapeout.transforms.utils` instead of `tapeout.transforms`?

chick:
Mostly I changed this because IntelliJ does not understand it well, and gives me the squigglies error indicators in all the files. But I do think it's cleared to have your package names to be within the src directory.

Regarding the utils package/not-a-package I was hesitant to change because add utils as a package would change the API of barstools tapeout. I don't know if people use the classes in there directly so I could just add a utils package to each file or the other alternative would be to move them up a directory so their packaged directories are correct as they stand.

I would like to clean this up. What do you think about which way to go?

abejgonzalez:
I think you can and `utils` as a new package (modify each file) and change the API. 

Also I think this `barstools.iocell.chisel/IOCell.scala` should be this instead `barstools/iocell/chisel/IOCell.scala`

abejgonzalez:
Closing since #97 covers this.

PR #96, opened 2021-02-04
Reformat code using scalafmt

chick:
Ran `scalafmt` over `iocells`, `macros`, and `tapeout`
This brings code into accordance with chisel standards
Easier to do now, before I do 3.4/3.5 refactor
This should be done after #95


abejgonzalez:
I think most of the formatting fixes look nice (i.e. adding spaces between operators, proper newlines between comments, etc). However, some things look bad IMO (sometimes `function(((` turns into 3 lines, etc)`. With that being said, this is just preference so if it is done automatically this should be fine. Once the other PR is addressed, Ill probably run CI on the combo of this one and the other to verify nothing magically broke.

chick:
There was much bike-shedding and gnashing of teeth over these formatting rules. What we have here was a hard won admittance by all that though no-one liked everything about it, at least these were automatic and made the larger code base consistent and overall easier to read.

abejgonzalez:
Closing this since #97 addresses this

PR #98, opened 2021-02-19
Fix deprecations 3

chick:
- Build out a stage for tapeout
  - Refactor annotation construction
  - Create a CLI handler
  - create a TapeOutStage
- Remove outputForm reference from EnumerateModules
- New GenerateTopAndHarness is fresh implmentation of Generate.scala
- GenerateTopSpec is a work in progress


tapeout/src/main/scala/barstools/tapeout/transforms/Generate.scala
```diff
@@ -1,274 +1,274 @@
-package barstools.tapeout.transforms
```
abejgonzalez, line 1: Can we just delete this file now?

chick: Deleted

abejgonzalez (approved):
LGTM since it is passing Chipyard CI

PR #99, opened 2021-02-22
Macrocompiler should prioritize memories with no masks with DefaultCostMetric

jerryz123:
When choosing between two equivalently sized library memories, we should choose the one with no mask, if our target memory also has no mask.


harrisonliew (approved):
Just wondering - is the relative cost of mask (0.001 fixed, feels arbitrary) vs. number of instances (0.05x) always the case?

jerryz123:
maskPenalty will always be less than costPerInst, which is by design. The default cost should still prioritize minimizing wasted bits, before trying to avoid unnecessary mask support.

PR #102, opened 2021-06-28
Use github actions for testing

chick:
- Stripped down version from treadle
- commented out several sections


chick:
Will try this again under another PR

PR #104, opened 2021-07-12
Restore proper naming of harness annotation file

jwright6323:
Somewhere along the way the usage of `HarnessAnnoOutAnnotation` was dropped in the `GenerateTopAndHarness` transform, which caused it to ignore the specified output annotation path and write to an incorrect one. This is the fix.


jerryz123 (approved):
LGTM

jwright6323:
Need a merge commit on this one to avoid breaking chipyard (mea culpa)

PR #105, opened 2021-07-12
Fix InstanceTarget annotations in ReParentCircuit

jwright6323:
Currently `ReParentCircuit` doesn't change any `InstanceTarget` annotations at all, so all annotations on `InstanceTarget`s are pruned when copying the `CircuitState`. This change looks for any `InstanceTarget` that passes through the new parent and re-parents them, along with changing the circuit string for all `InstanceTarget` annotations.


chick (approved):
This looks fine to me. 
It would be nice if there was a test that exercises the new code.

Minor quibble. Stylistically, the parens are not needed when you have braces in `.map` 
`.map({...})` can just be `.map {...}`

jwright6323:
Thanks @chick . I think in general barstools needs a lot of tests...

PR #106, opened 2021-07-13
Fix harness .fir file output location

jwright6323:
Another output file seems to have been dropped recently...

### DIFF
--- a/tapeout/src/main/scala/barstools/tapeout/transforms/GenerateTopAndHarness.scala
+++ b/tapeout/src/main/scala/barstools/tapeout/transforms/GenerateTopAndHarness.scala
@@ -15,6 +15,7 @@ import logger.LazyLogging
 private class GenerateTopAndHarness(annotations: AnnotationSeq) extends LazyLogging {
   val synTop: Option[String] = annotations.collectFirst { case SynTopAnnotation(s) => s }
   val topFir: Option[String] = annotations.collectFirst { case TopFirAnnotation(s) => s }
+  val harnessFir: Option[String] = annotations.collectFirst { case HarnessFirAnnotation(s) => s }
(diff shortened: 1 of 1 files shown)


jwright6323:
@jerryz123 @chick This and #105 seem to have trouble with `sbt test`. I'm able to replicate it locally but couldn't fix it. Any ideas?

tapeout/src/main/scala/barstools/tapeout/transforms/GenerateTopAndHarness.scala
```diff
@@ -111,7 +112,7 @@ private class GenerateTopAndHarness(annotations: AnnotationSeq) extends LazyLogg
     val annos = new FirrtlStage().execute(Array.empty, generatorAnnotations)
     annos.collectFirst { case FirrtlCircuitAnnotation(circuit) => circuit } match {
       case Some(circuit) =>
-        dump(circuit, annos, topFir, harnessAnnoOut)
+        dump(circuit, annos, harnessFir, harnessAnnoOut)
```
chick, line 115: This looks like it will be a change in behavior, are people using this already using HarnessFirAnnotation?
Should harnessFir fall back to TopFirAnnotation if HarnessFirAnnotation is not present?

chick (commented):
Can you respond to questions

tapeout/src/main/scala/barstools/tapeout/transforms/GenerateTopAndHarness.scala
jwright6323: This is a bug fix; the second pass is actually overwriting topFir. It should not fall back.

chick (approved):
ok, lgtm

PR #109, opened 2021-08-10
Unify barstools into single src tree that matches package structure

chick:
This is a pretty massive PR that does several things
- Slurps in `plsi-mdf` into this repo. 
- Has a single top level `src/` dir that includes all
- Fixes a lot of tests that relied on each other's side effects
This will make maintaining this repo easier as tests are unified and easier to run.
There are still some cleanup things to do. 
In particular tests that call directly into a `StageMain` subclass's `main` method while capturing stdout can create some very difficult to debug tests.


edwardcwang (approved):
Generally seems fine to me, but the others should take a look too.

ekiwi (approved):
LGTM

jerryz123:
Can we bump barstools in chipyard as well with this?

chick:
@jerryz123 
Also this is built against chisel 3.5 and iotesters 2.5 and that may require a bit of tinkering to chipyard
I think there might be a tweak or two to regarding with build.sbt and packages
I'll try it soon

harrisonliew (approved):

PR #112, opened 2021-08-17
Fix all warnings in barstool.macros._

chick:
- Fixed up all warnings in barstools macros package
- mostly public method return types
- removed lot's of extraneous parens and braces
- Made code cleaner using more explicit macros
- Fixed warnings in 2.13 that will likely turn into errors in future

### DIFF
--- a/.github/workflows/run-ci.yml
+++ b/.github/workflows/run-ci.yml
@@ -29,6 +29,10 @@ jobs:
     name: Documentation and formatting
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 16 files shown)


chick:
I'm kind of hoping for a rubber stamp here. This is all just stylistic and convention changes that will make it easier for me to move forward with some MacroCompiler enhancements. A lot of these changes should help with impending scala changes.

ekiwi (approved):
Looks good in general, but there are two tests that are failing.

PR #113, opened 2021-08-24
avoid collision with new firrtl.Utils.and

timsnyder-siv:
explicitly list imports from firrtl.Utils because barstools.macros.Utils
defines an 'and' function that will conflict with the newly added firrtl.Utils.and

### DIFF
--- a/macros/src/main/scala/barstools/macros/MacroCompiler.scala
+++ b/macros/src/main/scala/barstools/macros/MacroCompiler.scala
@@ -8,7 +8,7 @@
 package barstools.macros
 
 import barstools.macros.Utils._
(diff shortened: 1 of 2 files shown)


ekiwi:
I can't find the exact PR, but I am pretty sure @chick recently fixed this.

timsnyder-siv:
@chick we're a bit behind master but I wanted to push this change that @jackkoenig says we need to make before `firrtl.Utils.and` is added.  

@ekiwi , I'll look for it and close this PR unmerged if it's already handled in master.

timsnyder-siv:
Looks like it was addressed as a part of  #109

PR #115, opened 2021-09-27
Clean up the assembly process

chick:
Assembly of a jar using main from MacroCompiler is fixed.

I don't see anywhere where this is used. So I'm not sure what may be affected by the placement
of the generated jar in ./utils/bin.

- Bump version so it's in step with Chisel3 3.5
- Avoid error on duplicate classes
- put jar in utils/bin, the same way that firrtl does assembly
- add calling script ./utils/bin/macro_compiler.sh

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,7 +7,7 @@ val defaultVersions = Map(
 
 lazy val commonSettings = Seq(
   organization := "edu.berkeley.cs",
-  version := "0.4-SNAPSHOT",
+  version := "0.5-SNAPSHOT",
   scalaVersion := "2.12.13",
   crossScalaVersions := Seq("2.12.13", "2.13.6"),
   scalacOptions := Seq("-deprecation", "-feature", "-language:reflectiveCalls"),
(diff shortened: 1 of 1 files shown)


abejgonzalez (commented):
Has this been bumped and tested w/ Chipyard?

abejgonzalez:
Is there supposed to be a script added (according to the PR description)?

abejgonzalez:
I don't think CY currently uses this JAR, so is the point to make it this easier to use when we move to a published version of this repo?

chick:
I haven't bumped and tested in chipyard. I'll give that a try, but barstools master is now on chisel 3.5-snapshot, I think that might be a problem

chick:
Not sure about the script added. I was just flushing the assembly out a bit with the firrtl convention for providing a script to run the assembled jar. I don't think CY uses this at all. The assembly process looks like @edwardcwang added it. I think assembly jars are nice because they spin up way faster than sbt, so ultimately it would be good to get chipyard to use them where feasible.

Issue #117, opened 2022-02-02
tapeout project is overwriting Chipyard build.sbt settings


tymcauley:
I originally filed this issue in Chipyard, and there's some useful context there: https://github.com/ucb-bar/chipyard/issues/1103

It looks like the definition of a `tapeout` project (the `lazy val tapeout = (project ...)` line) in this repo's `build.sbt` file causes conflicts with parent repos when barstools is used in a multi-project SBT setup (like Chipyard). In the case of Chipyard, all of Chipyard's `tapeout` project settings are getting overwritten by the `tapeout`  project settings in this repo's `build.sbt` file, so users can't customize the `tapeout` project definition in Chipyard. For my use-case, I'd like to add a dependency on another Chipyard project which contains some FIRRTL transforms. However, those cross-project dependencies get removed by the `tapeout` project definition in this repo's `build.sbt` file.

I'm not quite familiar enough with this project to know if this will cause issues, but it appears that removing the `tapeout` project from this repo's `build.sbt` file and making all of the settings unscoped resolves this conflict. For instance, the `build.sbt` file for this project could look something like this:

```scala
val defaultVersions = Map(
  "chisel3" -> "3.5-SNAPSHOT",
  "chisel-iotesters" -> "2.5-SNAPSHOT"
)

organization := "edu.berkeley.cs"
version := "0.4-SNAPSHOT"
name := "tapeout"
scalaVersion := "2.12.13"
fork := true
scalacOptions := Seq("-deprecation", "-feature", "-language:reflectiveCalls")
Test / scalacOptions ++= Seq("-language:reflectiveCalls")
mainClass := Some("barstools.macros.MacroCompiler")
libraryDependencies ++= Seq("chisel3","chisel-iotesters").map {
  dep: String => "edu.berkeley.cs" %% dep % sys.props.getOrElse(dep + "Version", defaultVersions(dep))
}
libraryDependencies ++= Seq(
  "com.typesafe.play" %% "play-json" % "2.9.2",
  "org.scalatest" %% "scalatest" % "3.2.9" % "test",
  "org.apache.logging.log4j" % "log4j-api" % "2.11.2",
  "org.apache.logging.log4j" % "log4j-core" % "2.11.2"
)
resolvers ++= Seq(
  Resolver.sonatypeRepo("snapshots"),
  Resolver.sonatypeRepo("releases"),
  Resolver.mavenLocal
)
```

[...]

PR #118, opened 2022-02-02
Remove sbt subproject "tapeout"

tymcauley:
Now barstools should be easier to compose in multi-project sbt projects, such as Chipyard.

Resolves ucb-bar/barstools#117.

Differences between this setup and the existing one:

- Removed `enablePlugins(sbtassembly.AssemblyPlugin)`. I think that since `sbt-assembly` is in `project/plugins.sbt`, this shouldn't be required. I was able to run `sbt assembly`, so that seems fine.
- Added `name := "tapeout"`, so the project name is `tapeout` when you enter an `sbt` console.

Everything else _should_ be preserved from the existing setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ Barstools
 
 ![Test](https://github.com/freechipsproject/ucb-bar/barstools/Test/badge.svg)
 
-**Barstools** is a coolection of useful utilities for BAR projects
+**Barstools** is a collection of useful utilities for BAR projects
 
 Passes/Transforms that could be useful if added here:
 * Check that a module was de-duplicated. Useful for MIM CAD flows and currently done in python.
@@ -18,16 +18,13 @@ Example Usage:
 ```
 sbt
 > compile
-> project tapeout
 > runMain barstools.tapeout.transforms.GenerateTop -i <myfile>.fir -o <myfile>.v --syn-top <mysyntop> --harness-top <myharnesstop>
 ```
 Building the macro compiler JAR:
 ```
 $ sbt
 [...]
-[info] Set current project to root (in build file:/mnt/data/dev/barstools_pcad/)
-> project macros
-[info] Set current project to macros (in build file:/mnt/data/dev/barstools_pcad/)
+[info] Set current project to tapeout (in build file:/mnt/data/dev/barstools_pcad/)
 > assembly
 [...]
 [info] SHA-1: 77d4c759c825fd0ea93dfec26dbbb649f6cd5c89
--- a/build.sbt
+++ b/build.sbt
@@ -5,39 +5,26 @@ val defaultVersions = Map(
   "chisel-iotesters" -> "2.5-SNAPSHOT"
 )
 
-lazy val commonSettings = Seq(
-  organization := "edu.berkeley.cs",
-  version := "0.4-SNAPSHOT",
-  scalaVersion := "2.12.13",
-  crossScalaVersions := Seq("2.12.13", "2.13.6"),
-  scalacOptions := Seq("-deprecation", "-feature", "-language:reflectiveCalls"),
-  libraryDependencies ++= Seq("chisel3","chisel-iotesters").map {
-    dep: String => "edu.berkeley.cs" %% dep % sys.props.getOrElse(dep + "Version", defaultVersions(dep))
-  },
-  libraryDependencies ++= Seq(
-    "com.typesafe.play" %% "play-json" % "2.9.2",
-    "org.scalatest" %% "scalatest" % "3.2.9" % "test",
-    "org.apache.logging.log4j" % "log4j-api" % "2.11.2",
-    "org.apache.logging.log4j" % "log4j-core" % "2.11.2"
-  ),
-  resolvers ++= Seq(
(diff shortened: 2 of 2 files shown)


tymcauley:
I'm seeing these same errors on my local machine, both while using this branch and also on `master`. Could this be due to using the chisel/firrtl snapshot repos, and them having been updated?

ekiwi:
Not sure what is going on, but seems like it would be a good idea to just add the compiler plugin to the `build.sbt`.
like: https://github.com/freechipsproject/chisel-template/blob/6cbfcce017880c3834965ffbe13d69376ef4e450/build.sbt#L22

ekiwi:
Also, this project should be switched to either the `x.6-SNAPSHOT` or the `X.5.1` release. `X.5-SNAPSHOT` isn't current anymore.

tymcauley:
> Not sure what is going on, but seems like it would be a good idea to just add the compiler plugin to the `build.sbt`. like: https://github.com/freechipsproject/chisel-template/blob/6cbfcce017880c3834965ffbe13d69376ef4e450/build.sbt#L22

Yup, adding the compiler plugin fixed the test failures on my machine. I'll update that.

tymcauley:
> Also, this project should be switched to either the `x.6-SNAPSHOT` or the `X.5.1` release. `X.5-SNAPSHOT` isn't current anymore.

@abejgonzalez do you have a preference for which version of chisel/iotesters I should use?

ekiwi:
> @abejgonzalez do you have a preference for which version of chisel/iotesters I should use?

we can probably do this in a separate PR

abejgonzalez:
I think we should use Chisel 3.5.1 (I'm not sure if it is released yet - last I heard was that by the end of the week it would be released - @ekiwi any updates?). I also think we can update it here (when ready) so that we only have 1 PR to deal with (and bump in Chipyard).

tymcauley:
> I think we should use Chisel 3.5.1 (I'm not sure if it is released yet - last I heard was that by the end of the week it would be released - @ekiwi any updates?). I also think we can update it here (when ready) so that we only have 1 PR to deal with (and bump in Chipyard).

Ya, looks like it's not released yet.

ekiwi:
> Ya, looks like it's not released yet.

I think if @abejgonzalez wants the Chisel update to be part of this PR, just changing the version to `3.5.0` is the right way to go.

tymcauley:
> > Ya, looks like it's not released yet.
> 
> I think if @abejgonzalez wants the Chisel update to be part of this PR, just changing the version to `3.5.0` is the right way to go.

Should we also use 2.5.0 for chisel-iotesters?

ekiwi:
> Should we also use 2.5.0 for chisel-iotesters?

That would be a requirement. Versions have to be lined up: https://www.chisel-lang.org/chisel3/docs/appendix/versioning.html

abejgonzalez:
This PR will most likely be merged with Chipyard's 3.5.1 bump so I rather just wait for that to get released (and have this PR get updated to 3.5.1)

tymcauley:
> This PR will most likely be merged with Chipyard's 3.5.1 bump so I rather just wait for that to get released (and have this PR get updated to 3.5.1) - In theory shouldn't be too long from now

Sounds good, I'll be on the lookout for that release, and I'll bump this branch once that happens.

davidbiancolin:
In general, removing the val commonSettings and promoting everything into default scope (as you have done here) makes it much easier to do multi-project builds. Using freshProject was a workaround for not being able to override settings in sub-subprojects due to the use of the commonSettings pattern.

tymcauley:
> In general, removing the val commonSettings and promoting everything into default scope (as you have done here) makes it much easier to do multi-project builds in SBT. Using freshProject was a workaround for not being able to override settings in sub-subprojects due to the use of the commonSettings pattern.

Nice! I'm glad I stumbled into a good solution for this :)

Originally I noticed that this project's `build.sbt` was set up very differently from the "normal" way that Chipyard recommends adding in subprojects: https://chipyard.readthedocs.io/en/latest/Customization/Custom-Chisel.html

abejgonzalez:
Ok. Double-checking this works with the Chisel 3.5.1 CY branch (using the `remove-sbt-subproject-1` branch in this repo). If that works, I'm fine merging this.

tymcauley:
> Ok. Double-checking this works with the Chisel 3.5.1 CY branch (using the `remove-sbt-subproject-1` branch in this repo). If that works, I'm fine merging this (let me know if this isn't ready to merge just in case - I'll probably just click the button when ready).

Yup, as far as I'm concerned this branch is ready to merge. Thanks!

tymcauley:
Seems like that PR in CY is happy with this branch. You're not waiting on anything from me, right @abejgonzalez?

abejgonzalez:
Nope. We are just confirming things are fine on the FireSim end then we should release soon (hopefully by EOD tomorrow)

tymcauley:
> Nope. We are just confirming things are fine on the FireSim end then we should release soon (hopefully by EOD tomorrow)

Sounds great, thanks!

abejgonzalez (approved):
LGTM

PR #123, opened 2022-12-24
CIRCT Integration

joey0320:
Look at (ucb-bar/chipyard#1239).
Remove SFC passes (will be replaced by FIRTOOL).


src/main/scala/barstools/tapeout/transforms/GenerateTopAndHarness.scala
```diff
@@ -7,118 +7,78 @@ import firrtl.ir._
 import firrtl.options.{Dependency, InputAnnotationFileAnnotation, StageMain}
 import firrtl.passes.memlib.ReplSeqMemAnnotation
 import firrtl.stage.{FirrtlCircuitAnnotation, FirrtlStage, OutputFileAnnotation, RunFirrtlTransformAnnotation}
+import firrtl.passes.{ConvertFixedToSInt}
```
abejgonzalez, line 10: Remove?

src/main/scala/barstools/tapeout/transforms/GenerateTopAndHarness.scala
```diff
@@ -7,118 +7,78 @@ import firrtl.ir._
...
-  val topAnnos = synTop.map(st => ReParentCircuitAnnotation(rootCircuitTarget.module(st))) ++
-    topDotfOut.map(BlackBoxResourceFileNameAnno)
-
   // Dump firrtl and annotation files
   protected def dump(
     circuit:     Circuit,
-    annotations: AnnotationSeq,
-    firFile:     Option[String],
-    annoFile:    Option[String]
+    annotations: AnnotationSeq
```
abejgonzalez, line 28: I would rename this function as `dumpAnnos` and remove the `circuit` parameter.

src/main/scala/barstools/tapeout/transforms/GenerateTopAndHarness.scala
```diff
@@ -7,118 +7,78 @@ import firrtl.ir._
...
-        harnessTop.map(ht => ModuleNameSuffixAnnotation(rootCircuitTarget, s"_in${ht}")) ++
-        synTop.map(st => ConvertToExtModAnnotation(rootCircuitTarget.module(st))) ++
-        Seq(
-          LinkExtModulesAnnotation(topExtModules),
-          RunFirrtlTransformAnnotation(Dependency[ConvertToExtMod]),
-          RunFirrtlTransformAnnotation(Dependency[RemoveUnusedModules]),
-          RunFirrtlTransformAnnotation(Dependency[AvoidExtModuleCollisions]),
-          RunFirrtlTransformAnnotation(Dependency[AddSuffixToModuleNames])
-        )
+    executeTop()
```
abejgonzalez, line 60: Is there a need to have `executeTop` and the `TopAndHarness` version of the functions? In the past, we needed this split since we separated out the FIRRTL runs on separate modules. Now that new flow supports each module being in a different file this isn't needed. I would just have a single entrypoint (single execute function) and remove the naming around top and harness (since this distinction doesn't matter anymore).

src/main/scala/barstools/tapeout/transforms/GenerateTopAndHarness.scala
```diff
@@ -7,118 +7,78 @@ import firrtl.ir._
...
 
     // For harness run, change some firrtlOptions (below) for harness phase
     // customTransforms: setup harness transforms, add AvoidExtModuleCollisions
     // outputFileNameOverride: change to harnessOutput
     // conf file must change to harnessConf by mapping annotations
-
```
abejgonzalez, line 103: I would also do a pass on all the comments since some don't apply anymore ^.

src/main/scala/barstools/tapeout/transforms/GenerateTopAndHarness.scala
abejgonzalez: I also generally don't like functions that read global parameters (instead of using function parameters). For now this is fine, but could you add a comment that this function reads the global `outAnno`.

abejgonzalez (changes requested):

src/main/scala/barstools/tapeout/transforms/GenerateTopAndHarness.scala
joey0320: Since we are using `mems.conf` generated by the `SFC` for the `Chisel` -> `LowFIRRTL` -> `Verilog` case, I think we need both `executeTop` and `executeTopAndHarness` to generate both the `top.mems.conf` and `harness.mems.conf`. https://github.com/ucb-bar/barstools/blob/13e2bb92ab7360f0ce3b2ff4d27b3024736111fb/src/main/scala/barstools/tapeout/transforms/GenerateTopAndHarness.scala#L69 this code seems to guide the `SFC` to generate `harness.mems.conf` instead of `top.mems.conf`. Would there be a cleaner way to generate both `top.mems.conf` & `harness.mems.conf` without maintaining `executeTop` and `executeTopAndHarness`?

abejgonzalez (approved):
LGTM

PR #125, opened 2023-02-06
Run RemoveValidIf pass for updated CIRCT

abejgonzalez:
Adds an extra flag that allows the compiler to run extra LowFirrtl passes needed for CIRCT to pass (i.e. run the RemoveValidIf pass).


abejgonzalez:
Another way to do this is to add a `RunTransformAnnotation` to the `.json` that we emit in CY that specifies the xform to run (i.e. the `ExtraPassesTransform`). That will avoid us having to add an extra compiler flag that is redundant.

joey0320:
I agree that having duplicate compile flags isn't clean. Are we only performing the `ExtraPassesTransform` when only using the `ENABLE_CUSTOM_FIRRTL_PASS` flag?

abejgonzalez:
> I agree that having duplicate compile flags isn't clean. Are we only performing the `ExtraPassesTransform` when only using the `ENABLE_CUSTOM_FIRRTL_PASS` flag?

Correct. We are only running this when dumping out LowFIRRTL because CIRCT no longer supports `validif` statements (which are emitted in SFC LowFIRRTL)

joey0320:
I see. I can try deduplicating this flag later. In the meanwhile, we can also just merge this in.

abejgonzalez:
Let's merge this in when the `new-hammer` work gets merged in.

joey0320:
okey-dokey

harrisonliew (approved):
Verified that this works. Can we merge this for ucb-bar/chipyard#1324?

PR #129, opened 2023-03-18
Add name to IOCell definition

kevindna:
Added variable to i_name in IOCell to contain name of signal to IOBinders

### DIFF
--- a/src/main/scala/barstools/iocell/chisel/IOCell.scala
+++ b/src/main/scala/barstools/iocell/chisel/IOCell.scala
@@ -60,14 +60,14 @@ class DigitalInIOCellBundle extends Bundle {
 }
 
 trait IOCell extends BaseModule {
-	var iocell_name : Option[String] = None
+  var iocell_name: Option[String] = None
 
-  /** Set IOCell name 
-   * @param s Proposed name for the IOCell
-   *
-   * @return An inherited IOCell with given the proposed name
-   */
-  def suggestName(s: String) : this.type = {
+  /** Set IOCell name
+    * @param s Proposed name for the IOCell
+    *
+    * @return An inherited IOCell with given the proposed name
+    */
+  def suggestName(s: String): this.type = {
     iocell_name = Some(s)
     super.suggestName(s)
   }
@@ -124,8 +124,6 @@ case class GenericIOCellParams() extends IOCellTypeParams {
   def output() = Module(new GenericDigitalOutIOCell)
 }
 
-
-
 object IOCell {
 
   /** From within a RawModule or MultiIOModule context, generate new module IOs from a given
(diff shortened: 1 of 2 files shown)


src/main/scala/barstools/iocell/chisel/IOCell.scala
```diff
@@ -59,7 +59,9 @@ class DigitalInIOCellBundle extends Bundle {
   val ie = Input(Bool())
 }
 
-trait IOCell extends BaseModule
+trait IOCell extends BaseModule {
+  var i_name: String
```
jerryz123, line 63: Can you make this name more descriptive? Also try
```
var iocell_name: Option[String] = None
override def suggestName(s: String): IOCell = {
  iocell_name = Some(s)
  super.suggestName(s)
}
```

If you do it this way, you don't have to set a default in the classes which inherit this trait. 
Also you don't have to do another suggestName in generateIOCell

src/main/scala/barstools/iocell/chisel/IOCell.scala
```diff
@@ -112,7 +118,11 @@ case class GenericIOCellParams() extends IOCellTypeParams {
   def output() = Module(new GenericDigitalOutIOCell)
 }
 
-object IOCell {
+trait IOCellName {
+  var i_name: String
```
jerryz123, line 122: This shouldn't be necessary

src/main/scala/barstools/iocell/chisel/IOCell.scala
```diff
@@ -112,7 +118,11 @@ case class GenericIOCellParams() extends IOCellTypeParams {
...
+}
+
+object IOCell extends IOCellName {
```
jerryz123, line 126: This shouldn't be necessary

src/main/scala/barstools/iocell/chisel/IOCell.scala
```diff
@@ -156,18 +168,25 @@ object IOCell {
     )(coreSignal:   T,
       padSignal:    T
     ): Seq[IOCell] = {
+      print("Suggested names:  " + name + "    ")
```
jerryz123, line 171: Remove this print statement? Or do you think its helpful to have this here?

If you think it is helpful, we should make it more descriptive: 
"Generating IOCell [IOCellType] named [name]"

jerryz123 (changes requested):

src/main/scala/barstools/iocell/chisel/IOCell.scala
kevindna: not helpful. just missed this one. will remove

src/main/scala/barstools/iocell/chisel/IOCell.scala
jerryz123: Yeah. It minimizes the amount of code which has to change, any thing which tries to name an IOCell with suggestName will also set the stored io_cell name with this change

jerryz123 (approved):

PR #131, opened 2023-05-31
Move iocells to separate project root

ethanwu10:
Currently Chipyard is including `barstools.iocell` as a separate project from the rest of barstools, and it does this by creating a new project rooted at `src` with its source directory rooted at `src/main/scala/barstools/iocell`. This causes problems for IDEs like IntelliJ which do not support having one source root from one project nested inside another source root from a different project.

By breaking out `barstools.iocell` into a separate project root, this should cause IDEs and similar tools to better understand the project structure, and makes it so that `iocell` is not in two projects at the same time.

This PR is currently still a draft because as it stands right now, the CI in *this* repo will ignore the `iocells` project, since it's not referenced by the `build.sbt` of this repo.

**Linked PRs**
chipyard: ucb-bar/chipyard#1491


jerryz123:
I think omitting compile of the iocell project is worth the improved IDE compatibility. I'm going to move forwards with merging this

jerryz123 (approved):

